PR #3, opened 2017-01-26
Additional set up detail; version bump?

chrisaldrich:
Prompted by query at https://wordpress.org/support/topic/your-have-entered-a-valid-domain-but-you-have-no-account-on-this-blog/ 

Might also be time for a quick version bump to indicate it still works and is supported? The official version in the WordPress.org repository [https://wordpress.org/plugins/indieauth/] is just behind the Github one and indicates the plugin "may be out of date" which never helps for adoption.


pfefferle:
thanks

PR #9, opened 2018-01-01
Complete Rewrite of Plugin

dshanske:
I know it is easier to digest in chunk sized bits, but this is a major rewrite of the plugin, after which my intention is to go smaller for future enhancements. This...

1. Rewrites the README a bit
1. Adds support for login using Indieauth Bearer Tokens, which actually now makes this a supported authentication plugin for the WordPress REST API.
1. Adds using the author profile as your URL
1. Add settings to disable the login form addition and to use endpoints to something other than Indieauth.com
1. Supports setting Indieauth Scope as a global variable for future use as WordPress has no built in scope options
1. Updates functionality to the Indieauth specification at indieauth.net
1. Adds a state parameter being passed, even though it isn't used as required for compliance.

I drew some inspiration from how @snarfed set up IndieAuth logins in Micropub, as well as some proof of concept OAuth2 plugins for WordPress, and read through @aaronpk 's OAuth 2.0 book.

I don't think I'll stop here.

### DIFF
--- a/includes/class-indieauth-authenticate.php
+++ b/includes/class-indieauth-authenticate.php
@@ -1,0 +1,271 @@
+<?php
+/**
+ * Authentication class
+ * Helper functions for extracting tokens from the WP-API team Oauth2 plugin
+ */
+class IndieAuth_Authenticate {
+
+	public function __construct() {
+		add_filter( 'determine_current_user', array( $this, 'determine_current_user' ), 11 );
+		add_filter( 'rest_authentication_errors', array( $this, 'rest_authentication_errors' ) );
+		add_action( 'authenticate', array( $this, 'authenticate' ) );
+	}
+
+	/**
+	 * Report our errors, if we have any.
+	 *
+	 * Attached to the rest_authentication_errors filter. Passes through existing
+	 * errors registered on the filter.
+	 *
+	 * @param WP_Error|null Current error, or null.
+	 *
+	 * @return WP_Error|null Error if one is set, otherwise null.
+	 */
+	public function rest_authentication_errors( $error = null ) {
+		if ( ! empty( $error ) ) {
+			return $error;
+		}
+		global $indieauth_error;
+		return $indieauth_error;
+	}
+
+	public function determine_current_user( $user_id ) {
+		$token = $this->get_provided_token();
+		if ( ! $token ) {
+			return $user_id;
+		}
+		$args     = array(
+			'headers' => array(
+				'Accept'        => 'application/json',
+				'Authorization' => 'Bearer ' . $token,
+			),
+		);
+		$response = wp_safe_remote_get( get_option( 'indieauth_token_endpoint' ), $args );
+		$code     = wp_remote_retrieve_response_code( $response );
+		$body     = wp_remote_retrieve_body( $response );
+
+		if ( 2 !== (int) ( $code / 100 ) ) {
+			global $indieauth_error;
+			$indieauth_error = new WP_Error(
+				'indieauth.invalid_access_token',
+				__( 'Supplied Token is Invalid', 'indieauth' ),
+				array(
+					'status'   => $code,
+					'response' => $body,
+				)
+			);
+			return $user_id;
+		}
+		$params = json_decode( $body, true );
+		global $indieauth_scopes;
+		$indieauth_scopes = explode( ' ', $params['scope'] );
+		$me               = $params['me'];
+		$user             = $this->get_user_by_identifier( $me );
+		if ( $user ) {
+			return $user->ID;
+		}
+
+		return $user_id;
+	}
+
+	/**
+	 * Authenticate user to WordPress using IndieAuth.
+	 *
+	 * @action: authenticate
+	 * @param mixed $user authenticated user object, or WP_Error or null
+	 * @return mixed authenticated user object, or WP_Error or null
+	 */
+	public function authenticate( $user ) {
+		$args        = array(
+			'headers' => array(
+				'Accept' => 'application/json',
+			),
+		);
+		$redirect_to = array_key_exists( 'redirect_to', $_REQUEST ) ? $_REQUEST['redirect_to'] : null;
+		if ( array_key_exists( 'indieauth_identifier', $_POST ) && $_POST['indieauth_identifier'] ) {
+				$me    = rawurlencode( $_POST['indieauth_identifier'] );
+				$query = build_query(
+					array(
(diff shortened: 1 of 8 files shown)


includes/class-indieauth-authenticate.php
```diff
@@ -0,0 +1,252 @@
...
+	 *
+	 * @param WP_Error|null Current error, or null.
+	 *
+	 * @return WP_Error|null Error if one is set, otherwise null.
+	 */
+	public function rest_authentication_errors( $error = null ) {
+		if ( ! empty( $error ) ) {
+			return $error;
+		}
+		global $indieauth_error;
```
pfefferle: why make it global?

includes/class-indieauth-authenticate.php
```diff
@@ -0,0 +1,252 @@
...
+				__( 'Supplied Token is Invalid', 'indieauth' ),
+				array(
+					'status'   => $code,
+					'response' => $body,
+				)
+			);
+			return $user_id;
+		}
+		$params = json_decode( $body, true );
+		global $indieauth_scopes;
```
pfefferle: why global?

includes/class-indieauth-authenticate.php
```diff
@@ -0,0 +1,252 @@
...
+
+	/**
+	 * Authenticate user to WordPress using IndieAuth.
+	 *
+	 * @action: authenticate
+	 * @param mixed $user authenticated user object, or WP_Error or null
+	 * @return mixed authenticated user object, or WP_Error or null
+	 */
+	public function authenticate( $user ) {
+		$args     = array(
```
pfefferle: can you have a look at the strange whitespaces

includes/class-indieauth-authenticate.php
```diff
@@ -0,0 +1,252 @@
...
+			'headers' => array(
+				'Accept'        => 'application/json'
```
pfefferle: also whitespaces

includes/class-indieauth-authenticate.php
```diff
@@ -0,0 +1,252 @@
...
+		// check result
+		if ( ! empty( $user_query->results ) ) {
+			return $user_query->results[0];
+		}
+
+		// Check author page
+		global $wp_rewrite;
+		$link = $wp_rewrite->get_author_permastruct();
+		if ( empty( $link ) ) {
+			$login = str_replace( home_url( '/' ) . '?author=', '', $identifier );
```
pfefferle: is it really home URL? not site URL?

pfefferle:
tested it with my blog and it does not work and I got no error message at all.

templates/indieauth-login-form.php
```diff
@@ -0,0 +1,6 @@
+<p style="margin-bottom: 8px;">
+	<label for="indieauth_identifier"><?php  __( 'Or login with your Domain', 'indieauth' ); ?> <br />
```
pfefferle: should be `_e()` instead!

templates/indieauth-login-form.php
```diff
@@ -0,0 +1,6 @@
...
+	<input type="text" name="indieauth_identifier" placeholder="your-domain.com" id="indieauth_identifier" class="input indieauth_identifier" value="" /></label>
+	<a href="https://indieauth.net/" target="_blank"><?php __( 'Learn about IndieAuth', 'indieauth' ); ?></a>
```
pfefferle: should be `_e()` instead!

pfefferle (changes requested):

dshanske:
Odd about it not working. Is this your development site, because I did switch to safe remote get and that if you remember from last time hates development sites

includes/class-indieauth-authenticate.php
dshanske: According to get_author_post_url

pfefferle:
I have it installed on my main blog

pfefferle:
but even if it is an issue, I miss an error message

includes/class-indieauth-authenticate.php
dshanske: There's no other way to surface the error. I was stumped about that till I looked at the REST API teams example Oauth plugin.

dshanske:
I will work on figuring that out. Did the old version work for you?

pfefferle:
sure

pfefferle:
at least it showed an error if I used a wrong URL or had a typo

dshanske:
Odd. I am not sure where to start, but I will start testing and working on better error handling

dshanske:
I think this is worth getting right.

pfefferle:
I have a lot of redirects and stuck at `https://notiz.blog/wp-login.php?redirect_to=...&code=...&me=...&state=...` at the end.

dshanske:
I will start trying to surface errors more effectively.

dshanske:
I think it is because I switched from the old verification method you were using to the current one in the specification. I may not have adjusted the error handling appropriately.

pfefferle:
this option is empty `indieauth_authorization_endpoint`

pfefferle:
It works if I remove this check https://github.com/indieweb/wordpress-indieauth/pull/9/files#diff-839da1f39b21c2dad138c57211f89cbdR94

dshanske:
I see. I'm looking at that right now. It was the last thing I implemented as a state parameter is required. I'm fixing it now.

dshanske:
@pfefferle I made fixes. I switched to using a nonce, which is more appropriate for a state parameter anyway. Should work now.

dshanske:
Also made the other changes requested.

dshanske:
It's my own fault. I wanted to finish before 2018.

dshanske:
Either way, while this is the only time I intend to put such a big change, I do want to work on making this even better with smaller commits and refinements subsequently.

pfefferle:
no problem, I fully support the re-writing!

pfefferle:
nice, seems to work for me, but I couldn't test the IndieAuth native login yet, I have no test instance.

pfefferle (approved):

dshanske:
I used gimme a token to get a token manually and curl to test.

pfefferle:
Ok, thanks!

dshanske:
I will add that as a suggestion to the readme.

pfefferle:
We should start with a indieauth provider so that we do not have to 😉

dshanske:
Define provider?

dshanske:
Do you mean built in endpoint?

PR #11, opened 2018-01-03
Unit Tests

dshanske:
Each function is separated into a separate commit.

1. Adds PHPUnit and the setup for unit testing
1. Adds validation and sanitization for URLs
1. Splits the various queries to external servers into separate functions so these functions can be unit tested
1. Discovered an issue with the previously added author profile search. Fixed it and tested to ensure working to spec
1. Reduced scope of the Indieauth Error Variable. No way to do same for the scope variable. 
1. Adds the first unit test.


pfefferle (approved):
looks good

PR #12, opened 2018-01-03
added CI tools

pfefferle:


pfefferle:
Sure, but you are the specialist for that 😉

PR #16, opened 2018-01-16
Discovery

dshanske:
Add discovery functions and testing to Indieauth

### DIFF
--- a/includes/class-indieauth-authenticate.php
+++ b/includes/class-indieauth-authenticate.php
@@ -10,6 +10,19 @@ class IndieAuth_Authenticate {
 		add_filter( 'determine_current_user', array( $this, 'determine_current_user' ), 11 );
 		add_filter( 'rest_authentication_errors', array( $this, 'rest_authentication_errors' ) );
 		add_action( 'authenticate', array( $this, 'authenticate' ) );
+
+		add_action( 'send_headers', array( $this, 'http_header' ) );
+		add_action( 'wp_head', array( $this, 'html_header' ) );
+	}
+
+	public static function http_header() {
+		header( sprintf( 'Link: <%s>; rel="authorization_endpoint"', get_option( 'indieauth_authorization_endpoint' ) ), false );
+		header( sprintf( 'Link: <%s>; rel="token_endpoint"', get_option( 'indieauth_token_endpoint' ) ), false );
+	}
+
+	public static function html_header() {
+		printf( '<link rel="authorization_endpoint" href="%s" />' . PHP_EOL, get_option( 'indieauth_authorization_endpoint' ) );
+		printf( '<link rel="token_endpoint" href="%s" />' . PHP_EOL, get_option( 'indieauth_token_endpoint' ) );
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


dshanske:
@pfefferle Any chance you can review this today? Want to use it for part of my hack day project.

includes/class-indieauth-authenticate.php
```diff
@@ -10,6 +10,22 @@ public function __construct() {
 		add_filter( 'determine_current_user', array( $this, 'determine_current_user' ), 11 );
 		add_filter( 'rest_authentication_errors', array( $this, 'rest_authentication_errors' ) );
 		add_action( 'authenticate', array( $this, 'authenticate' ) );
+		// Remove Micropub adding same headers
+		if ( ! defined( 'MICROPUB_LOCAL_AUTH' ) ) {
```
pfefferle: IndieAuth is the main authority for authorization and authentication and MicroPub only implements a fallback, so it should be the other way around. IndieAuth should set a const and MicroPub should check it.

includes/functions.php
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if ( isset( $return['me'] ) ) {
+		$me = $return['me'];
+	}
+	// check link header
+	$links = wp_remote_retrieve_header( $response, 'link' );
+	if ( $links ) {
+		if ( is_array( $links ) ) {
+			foreach ( $links as $link ) {
+				if ( preg_match( '/<(.[^>]+)>;\s+rel\s?=\s?[\"\']?(http:\/\/)?authorization_endpoint?\/?[\"\']?/i', $link, $result ) ) {
```
pfefferle: this regex is copied from the webmention code... you do not need to check for `http` ;)

dshanske: It is, I am regex challenged

includes/class-indieauth-authenticate.php
snarfed: for whether micropub uses the indieauth plugin or not? sure, micropub can look for a constant that indieauth defines.

we still want micropub to be usable without the indieauth plugin, thought, and ideally still let users decide whether to use indieauth.com or wordpress login, so we'd probably keep this constant for when the indieauth plugin isn't also installed.

dshanske: We don't need a constant, just if class exists then.

pfefferle: @snarfed I am not suggesting to remove the `MICROPUB_LOCAL_AUTH` constant literally, but to move the check to the micropub plugin instead of the indieauth plugin.

dshanske: I will amend the PR

pfefferle: Another problem with the actual merge request: If `MICROPUB_LOCAL_AUTH` is defined with `0` it will add both headers.

dshanske:
@pfefferle I rebased and adjusted.

includes/functions.php
```diff
@@ -0,0 +1,99 @@
...
+			foreach ( $links as $link ) {
+				if ( preg_match( '/<(.[^>]+)>;\s+rel\s?=\s?[\"\']?authorization_endpoint?\/?[\"\']?/i', $link, $result ) ) {
+					$return['authorization_endpoint'] = WP_Http::make_absolute_url( $result[1], $me );
+				}
+				if ( preg_match( '/<(.[^>]+)>;\s+rel\s?=\s?[\"\']?token_endpoint?\/?[\"\']?/i', $link, $result ) ) {
+					$return['token_endpoint'] = WP_Http::make_absolute_url( $result[1], $me );
+				}
+			}
+		} else {
+			if ( preg_match( '/<(.[^>]+)>;\s+rel\s?=\s?[\"\']?authorization_endpoint?\/?[\"\']?/i', $links, $result ) ) {
```
pfefferle: still some copy&paste issues... you can remove this `?\/?` behind all the rel value... It was, to optional support `http://webmention.org` with a trailing slash...

pfefferle (changes requested):

dshanske:
Hopefully this satisfies.

pfefferle (approved):

dshanske:
This is going to be a conflict with the new token endpoint I wrote. Will have to resubmit.

PR #19, opened 2018-01-21
Add Token Endpoint and add Authorization Discovery

dshanske:
This adds the discovery stuff from previous with the token endpoint items for today. It makes a bunch of changes recommended by @aaronpk as well.

Will likely hide the option to use an external token endpoint in a future version, as it seems that most people would want the built-in.

### DIFF
--- a/includes/class-indieauth-authenticate.php
+++ b/includes/class-indieauth-authenticate.php
@@ -10,7 +10,20 @@ class IndieAuth_Authenticate {
 		add_filter( 'determine_current_user', array( $this, 'determine_current_user' ), 11 );
 		add_filter( 'rest_authentication_errors', array( $this, 'rest_authentication_errors' ) );
 		add_action( 'authenticate', array( $this, 'authenticate' ) );
+
+		add_action( 'send_headers', array( $this, 'http_header' ) );
+		add_action( 'wp_head', array( $this, 'html_header' ) );
 	}
+
+	public static function http_header() {
+		header( sprintf( 'Link: <%s>; rel="authorization_endpoint"', get_option( 'indieauth_authorization_endpoint' ) ), false );
+		header( sprintf( 'Link: <%s>; rel="token_endpoint"', get_option( 'indieauth_token_endpoint' ) ), false );
+	}
+	public static function html_header() {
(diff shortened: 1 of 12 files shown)


dshanske:
Because @aaronpk had an issue with using localhost in testing, I added the DEBUG file @pfefferle created for webmentions when he had the same problem. Really looking forward to getting this in.

includes/class-indieauth-authenticate.php
```diff
@@ -81,50 +100,78 @@ public function verify_access_token( $token ) {
...
+				'me'            => rawurlencode( $me ),
+				'redirect_uri'  => rawurlencode( $redirect_uri ),
+				'client_id'     => rawurlencode( home_url() ),
+				'state'         => $state,
+				'response_type' => 'id',
 			)
 		);
 		// redirect to authentication endpoint
-		wp_redirect( get_option( 'indieauth_authorization_endpoint' ) . '?' . $query );
+		wp_redirect( $authorization_endpoint . '?' . $query );
```
pfefferle: you should use a more solid method to add the query to the authorization endpoint. If hoster does not support `mod_rewrite`, `$authorization_endpoint` might already has query params.

includes/class-indieauth-token-endpoint.php
```diff
@@ -0,0 +1,229 @@
...
+ */
+
+class IndieAuth_Token_Endpoint {
+	public function __construct() {
+		add_action( 'rest_api_init', array( $this, 'register_routes' ) );
+
+	}
+
+
+		/**
```
pfefferle: indents!

includes/class-indieauth-token-ui.php
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+
+	/**
+	 * Generate Options Form.
+	 *
+	 * @access public
+	 */
+	public function options_form() {
+		echo '<div class="wrap">';
```
pfefferle: perhaps use a template, or at least use a more template style output (don't echo HTML tags)

functions.php
```diff
@@ -0,0 +1,155 @@
+<?php
```
pfefferle: remove this file, because it is a duplicate of `includes/functions.php`

dshanske:
@pfefferle I removed the accidental adding of a duplicate functions file, changed the build_query to add_query_args which is more redundant, and moved the form to a template. Should do the trick.

PR #20, opened 2018-04-04
Add Authorization Endpoint and More

dshanske:


aaronpk:
This is missing the default value set to the new built-in authorization endpoint at https://github.com/dshanske/wordpress-indieauth/blob/master/indieauth.php#L64

aaronpk:
Is [this](https://github.com/dshanske/wordpress-indieauth/blob/token/includes/class-indieauth-token-endpoint.php#L177) supposed to be `if ( $return )` instead of `$token`? In my case the set token method returned false but the API call still gave the client a token even though it was not stored in the database.

aaronpk:
* I have my test site set to "single-user mode" in the IndieWeb plugin. 
* There is no value written to the `user_url` column for my user.

When the `get_user_by_identifier` function runs, it gets all the way to the [query by login function](https://github.com/dshanske/wordpress-indieauth/blob/token/includes/functions.php#L149) and no results are returned.

Maybe this plugin should also check if there is only one user in wordpress and return that user as a fallback?

dshanske:
As discussed, the idea of using the Indieweb plugin's single_author_setting should do the trick there.

aaronpk:
Something is going wrong with the query string parameters between all the redirect steps.

after i click log in on my client:

`https://tiny.xyz.dev/wp-login.php?redirect_to=https%3A%2F%2Ftiny.xyz.dev%2Fwp-login.php%3Fredirect_to%3Dhttps%253A%252F%252Fmicropub.rocks.dev%252Fendpoints%252Fcallback%26reauth%3D1%26action%3Dindieauth%26client_id%3Dhttps%3A%2F%2Fmicropub.rocks.dev%2F%26state%3D7042033%26scope%3Dcreate+update+delete+undelete%26me%3Dhttps%3A%2F%2Ftiny.xyz.dev`

note that it has the correct `+` between the scope values here.

after i type my password, I get taken to this URL:

`https://tiny.xyz.dev/wp-login.php?redirect_to=https%3A%2F%2Fmicropub.rocks.dev%2Fendpoints%2Fcallback&reauth=1&action=indieauth&client_id=https://micropub.rocks.dev/&state=7042033&scope=createupdatedeleteundelete&me=https://tiny.xyz.dev`

all the spaces are gone from the scope parameter.

templates/indieauth-authorize-form.php
```diff
@@ -0,0 +1,53 @@
...
+		'state' => $state,
+	),
+	$redirect_uri
+);
+?>
+<div class="login-info">
+<?php echo get_avatar( $current_user->ID, '78' ); ?>
+<?php
+	printf(
+		'<p>' . __( 'The app <strong>%1$s</strong> would like to access your site, <strong>%2$s</strong> using the credentials of <strong>%3$s</strong>(%4$s).', 'indieauth' ) . '</p>',
```
aaronpk: There should be an extra space between `</strong>(`.

Also it would be nice to skip showing the `(%4$s)` part if that is the same value as `%3`, for example when the user has not set a display name so both strings are the username.

aaronpk:
The authorization screen could use a bit of UI cleanup.

screenshot 2018-04-04 11 57 53

with a little more spacing around things it could look like this:

screenshot 2018-04-04 12 04 11

aaronpk:
I've broken these out into new issues so they'll be easier to discuss:

* https://github.com/indieweb/wordpress-indieauth/issues/24
* https://github.com/indieweb/wordpress-indieauth/issues/23
* https://github.com/indieweb/wordpress-indieauth/issues/22
* https://github.com/indieweb/wordpress-indieauth/issues/21

includes/class-indieauth-authorization-endpoint.php
```diff
@@ -0,0 +1,148 @@
...
+		if ( 'code' !== $params['response_type'] && 'id' !== $params['response_type'] ) {
+			return new WP_Error( 'unsupported_response_type', __( 'Unsupported Response Type', 'indieauth' ), array( 'status' => 400 ) );
+		}
+		$url = wp_login_url( $params['redirect_uri'], true );
+		$url = add_query_arg(
+			array(
+				'action'    => 'indieauth',
+				'client_id' => $params['client_id'],
+				'state'     => $params['state'],
+				'scope'     => isset( $params['scope'] ) ? $params['scope'] : 'create update',
```
aaronpk: don't set the default `create update` scope if `response_type=id`

aaronpk (changes requested):
I've collected a bunch of issues in a milestone so they can be discussed easier: https://github.com/indieweb/wordpress-indieauth/milestone/1

aaronpk (approved):

pfefferle:
Sorry, but I can't get it to work...

pfefferle:
OK, it was a caching problem, seems to work now! Nice!

pfefferle:
I can use my Blog to login to Quill or to the IndieWeb Wiki, bit the login to WordPress does not work.

aaronpk:
It worked for me, but of course I had to set the authorization endpoint in the settings back to `https://indieauth.com/auth`.

If you have the authorization endpoint set to the built-in one, then it would be kind of stuck in a weird state. Which brings me to another point. How do I know what the URL for the built-in authorization endpoint is? Right now if I change it to indieauth.com, then I have no way of knowing what to change it to to make it go back. Blanking out the value just stores a blank value which breaks it. Can you make it revert to the default built-in URL if a blank value is saved?

pfefferle:
@dshanske I used my main blog (notiz.blog) to login to (pfefferle.org) and I stuck at a white page.

aaronpk:
@pfefferle what authorization server is it set to use in that case?

pfefferle:
@aaronpk the internal on, that @dshanske implemented...

aaronpk:
Yeah I don't think that makes any sense, since you'd have to be logged in before you can log in. It only makes sense to sign in to wordpress using your domain if you're using an external authorization server.

aaronpk:
I'm not really sure how to solve the UI around this. tbh I think it would make sense to pull out the "log in to your wordpress" feature into a separate plugin, preferably called RelMeAuth or IndieLogin (to match the upcoming indielogin.com that I'm working on), since logging in to wordpress with your domain isn't actually doing IndieAuth.

I'm worried that having both in the same plugin is going to continue to confuse people.

dshanske:
I can get behind a fork

dshanske:
The alternative is to have a button on the login form that links to a relmeauth form. I can code that.

aaronpk:
Not sure what you mean, how does that change anything?

pfefferle:
@aaronpk but why? I want to use may domain to login to different sites using the internal indieauth?

dshanske:
For the default setting, I think I should change the interface to a radio button for default.

dshanske:
@pfefferle You should be able to do that

dshanske:
I can go back and redesign the interface. But should it be merged first and a second PR?

pfefferle:
@aaronpk sure, I understand the difference between rel-me auth and indieauth and I see, that it does not make sense to login to notiz.blog using notiz.blog... but using notiz.blog to login to pfefferle.org should work and that isn't working at the moment...

aaronpk:
I don't see this IndieAuth plugin installed on pfefferle.org: https://pfefferle.org/wp-login.php There's only an OpenID option there. Can you clarify exactly what you are expecting to happen?

pfefferle:
I removed it after some failed tests...

aaronpk:
So to clarify, you installed this plugin on both sites, and you were trying to log in to one as the other? It would be nice to get more information about the error you're having so we can try to reproduce it.

pfefferle:
@aaronpk what I tried to do:

* I added `https://notiz.blog/` to the login field on pfefferle.org
* Browser redirects me to indieauth.com
* I clicked at the `IndieAuth` authentication
* Browser redirects me to notiz.blog with the authentication screen
* I accepted the authentication dialog
* Browser redirects me to `https://pfefferle.org/wp-login.php?redirect_to=https%3A%2F%2Fpfefferle.org%2Fwp-admin%2F&code=XXX` and stays blank.

I will investigate why it stops rendering and try to debug the code. Perhaps I can send a merge request later today.

aaronpk:
Which site was this plugin installed at? Both? And which site was configured to point to indieauth.com?

Technically we shouldn't need to end up on indieauth.com for this use case, though that might need to be a separate set of changes to the plugin.

pfefferle:
@aaronpk both sites have the plugin installed and both are set to the native endpoint.

pfefferle:
@aaronpk @dshanske I think I found one problem: https://github.com/indieweb/wordpress-indieauth/pull/20/files#diff-839da1f39b21c2dad138c57211f89cbdL129 uses an old options entry that is not used any more!

pfefferle:
I think it was hardcoded to indieauth.com in an old version and we have to replace it with the auto-dicovered authorisation endpoint of the used URL!

dshanske:
@pfefferle Tell me if this is the problem. You try to log into notiz.blog with pfefferle.org and you click both times, login with my domain... So, the first time, it redirects you to pfefferle.org and if you don't click login with your domain the second time it will show the authenticate page. By clicking the check box, you are telling pfefferle.org to log you in using indieauth...Creating a loop... I think that interface needs improvement.

pfefferle:
This is what I am trying...

apr -06-2018 19-00-25

aaronpk:
indieauth.com should never have been involved in that flow if both wordpress plugins are set to use the internal authorization endpoint. I think your idea that an old options entry is being used is right.

pfefferle:
Is the micropub plugin adding indieauth headers???

pfefferle:
@dshanske @aaronpk the micropub plugin adds the default header below the indieauth headers, so the discovery gets the false headers... Can we disable that in the micropub plugin?

After deactivation the micropub plugin it does not load indieauth.com but I still can not login, because I stuck at a white page at the end!

pfefferle:
I always get:

```
WP_OAuth_Response Object
(
    [links:protected] => Array
        (
        )

    [matched_route:protected] =>
    [matched_handler:protected] =>
    [data] => Array
        (
            [error] => invalid_grant
            [error_description] => Invalid authorization code
        )

    [headers] =>
    [status] => 400
)
```

pfefferle:
Btw. the Quill login works as expected...

dshanske:
The development version doesn't do that. Expect @snarfed to be pushing it soon to fix the post status bug.

dshanske:
I am going to merge it then work on improvements in a new PR.

pfefferle:
@dshanske why have you merged it? I still stuck at a white page! and the error handling does not really work well!

dshanske:
I am going to work on a new PR now to address the issues. Apologies for the confusion, I thought you said it was working. I misread.

dshanske:
I think I need to fix the form and the error handling.

dshanske:
@pfefferle Will have the new PR tonight

PR #35, opened 2018-04-06
Endpoint Fixes

dshanske:
This tries to address the Error Logging issue by adding a conversion back to WP_Error in the Oauth Error Response function and adds checks to convert back.

It also changes the settings options to show radio buttons for custom, built-in, or Indieauth.com.

The login form to login with your domain is no longer shared with the regular login form and is accessed at wp-login.php?action=indielogin . There is a setting to show a link to it on the regular login form.

### DIFF
--- a/includes/class-indieauth-authenticate.php
+++ b/includes/class-indieauth-authenticate.php
@@ -114,14 +114,17 @@ class IndieAuth_Authenticate {
 
 	}
 
-	public function verify_access_token( $token ) {
+	public function verify_access_token( $token, $endpoint = null ) {
+		if ( ! $endpoint ) {
+			$endpoint = get_option( 'indieauth_token_endpoint', rest_url( 'indieauth/1.0/token' ) );
+		}
 		$args     = array(
 			'headers' => array(
 				'Accept'        => 'application/json',
 				'Authorization' => 'Bearer ' . $token,
 			),
 		);
-		$response = wp_safe_remote_get( get_option( 'indieauth_token_endpoint', rest_url( 'indieauth/1.0/token' ) ), $args );
+		$response = wp_safe_remote_get( $endpoint, $args );
 		if ( is_oauth_error( $response ) ) {
 			return $response;
 		}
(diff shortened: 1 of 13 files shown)


dshanske:
@pfefferle Hope this resolves your issue.

pfefferle:
Cool, now the error reporting works, but I still can't login!

```Invalid authorization code```

**Comment by pfefferle** [comment]
@aaronpk @dshanske as far as I understand OAuth and IndieAuth, you have to get and verify the tokens from the URL you want to login with... This code seems to use the endpoint from the URL I want to be logged in... https://github.com/indieweb/wordpress-indieauth/blob/master/includes/class-indieauth-authenticate.php#L177

Am I wrong?

**Comment by pfefferle** [comment]
same here: https://github.com/indieweb/wordpress-indieauth/blob/master/includes/class-indieauth-authenticate.php#L95

**Comment by dshanske** [comment]
That's my fault. The verify code was written when I was upgrading the original function, and I must not have upgraded it. Will address.

**Comment by dshanske** [comment]
Reading this more closely. verify_access_token does not get called except when logging into the site. Since, at the moment, whatever is in the option is what will be in the header, changing that to discover the endpoint won't work.

File: templates/indieauth-login-form.php [hunk 1]
```diff
@@ -1,7 +1,26 @@
...
+<div class="login-info">
+<p><?php _e( 'Sign in with your domain', 'indieauth' ); ?></p>
+<br />
+	<input type="url" name="indieauth_identifier" placeholder="<?php _e( 'https://example.com', 'indieauth' ); ?>" />
+<br />
+</div>
+<p class="submit">
+<?php
+// Hook to allow adding to form
+	add_action( 'indieauth_login_form' );
```
aaronpk: this is missing a second parameter:

> Fatal error: Uncaught ArgumentCountError: Too few arguments to function add_action(), 1 passed

includes/class-indieauth-authenticate.php
```diff
@@ -114,14 +114,17 @@ public function determine_current_user( $user_id ) {
 
 	}
 
-	public function verify_access_token( $token ) {
+	public function verify_access_token( $token, $endpoint = null ) {
+		if ( ! $endpoint ) {
+			$endpoint = get_option( 'indieauth_token_endpoint', rest_url( 'indieauth/1.0/token' ) );
```
pfefferle: this fallback does not work, you never want to use the local endpoint to verify!

includes/class-indieauth-authenticate.php
```diff
@@ -188,23 +191,25 @@ public function authorization_redirect( $me, $redirect_uri ) {
 		wp_redirect( $query );
 	}
 
-	public static function verify_authorization_code( $code, $redirect_uri, $client_id = null ) {
-		if ( ! $client_id ) {
-			$client_id = home_url();
+	// $args must consist of redirect_uri, client_id, and code
+	public static function verify_authorization_code( $post_args, $endpoint = null ) {
+		if ( ! $endpoint ) {
+			$endpoint = get_option( 'indieauth_authorization_endpoint' );
```
pfefferle: same as above, the local endpoint should never be used!

includes/class-indieauth-authenticate.php
```diff
@@ -114,14 +114,17 @@ public function determine_current_user( $user_id ) {
 
 	}
 
-	public function verify_access_token( $token ) {
+	public function verify_access_token( $token, $endpoint = null ) {
```
pfefferle: where is it `$endpoint` set?

pfefferle (changes requested):

includes/class-indieauth-authenticate.php
dshanske: Working on that now.

pfefferle (approved):

templates/indieauth-login-form.php
pfefferle: fixed

PR #38, opened 2018-04-14
Tweaks

dshanske:
This PR does the following

* Uses @pfefferle code from webfinger to replace the original code to find an author URL from an author post URL if provided. 
* It rearranges the search order for a URL to ID match to home_url first, if so searches for the Indieweb single author property, then searches whether it is a single user site, then searches for author post URL, and finally it does the meta query to search for the URL in user_url.
* It adds a filter to automatically add the trailing slash to all user_urls to ensure that it is always there when inserted into the database
* It adds testing
* Fixes a typo, redirect_url instead of redirect_uri
* Changes the settings page to allow for two presets(local or indieauth), or a custom option
* Adds functions to get the endpoints, instead of directly calling for the option.
* If there is no user_url, it will return the author post url as the 'me' parameter.
* Removes requirement for parameters from the REST call and add check inside the function so it will return the OAuth Error Response instead of the WP Error response.
* Adds a new debug function that will show the Indieauth requests in the error log if WP_DEBUG is enabled, specifically blanking the code and token data.


pfefferle:
I changed the form a bit, what do you think?

dshanske:
I like it. What do you think of the other changes?

pfefferle:
looks good to me... but I didn't check the functionality...

pfefferle (approved):

dshanske:
What do we need to do to push this to wordpress.org?

pfefferle:
merge it and then I can make a tag!

pfefferle:
is the latest version tested? can I release the 2.0.0 version?

dshanske:
I am using it actively.

PR #44, opened 2018-04-28
Local Token

dshanske:
This moves token verification when the endpoint is local to directly verifying the token, rather than making an http call. It also changes scope and response to being added to a filter, which can be used in any client to pull this, as opposed to a global variable.


includes/class-indieauth-authenticate.php
```diff
@@ -18,6 +20,17 @@ public function __construct() {
 
 		add_action( 'send_headers', array( $this, 'http_header' ) );
 		add_action( 'wp_head', array( $this, 'html_header' ) );
+
+		add_filter( 'indieauth_scopes', array( $this, 'indieauth_scopes' ), 9 );
+		add_filter( 'indieauth_response', array( $this, 'indieauth_response' ), 9 );
+	}
+
+	public static function indieauth_scopes( $scopes ) {
```
pfefferle: I would prefer more descriptive function names like `get_...`

pfefferle (approved):

PR #47, opened 2018-05-04
UI Fixes

dshanske:
Hides token settings page if local endpoint not enabled and switches  from error on redirect_uri not matching to warning.


pfefferle (approved):

dshanske:
@pfefferle Would you be able to push this version to .org?

PR #50, opened 2018-05-14
local verification of authorization codes

### LINKED ISSUE #49: Verify authorization codes locally if using the local endpoints
It looks like the plugin still tries to discover the user's authorization server even when the plugin is configured to use local endpoints. This should be bypassed similar to token verification and the codes should be looked up locally instead of making an HTTP request to itself.

aaronpk:
if the plugin is configured for local endpoints, this verifies authorization codes locally instead of making an HTTP request to itself

closes #49

### DIFF
--- a/includes/class-indieauth-authenticate.php
+++ b/includes/class-indieauth-authenticate.php
@@ -245,11 +245,6 @@ class IndieAuth_Authenticate {
 	// $args must consist of redirect_uri, client_id, and code
 	public static function verify_remote_authorization_code( $post_args, $endpoint ) {
 
-		$option = get_option( 'indieauth_config' );
-		if ( 'local' === $option ) {
-
-		}
-
 		if ( ! wp_http_validate_url( $endpoint ) ) {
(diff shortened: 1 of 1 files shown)


includes/class-indieauth-authenticate.php
```diff
@@ -223,8 +223,33 @@ public function authorization_redirect( $me, $redirect_uri ) {
...
+		}
+		return $return;
+	}
+
+	// $args must consist of redirect_uri, client_id, and code
+	public static function verify_remote_authorization_code( $post_args, $endpoint ) {
+
+		$option = get_option( 'indieauth_config' );
+		if ( 'local' === $option ) {
+
```
pfefferle: why this empty check?

dshanske: @aaronpk Do you have any comment on the empty check? Want to try to merge this.

aaronpk: Whoops, look like I copied some extra code from somewhere.

PR #54, opened 2018-05-18
Headers

dshanske:
I added a more complete polyfill for getallheaders from another source, added a composer file, and checked for an authorization in an alternate location common for FastCGI installations.


pfefferle:
I would move the `getallheaders` function to the `functions.php`

pfefferle (approved):

PR #67, opened 2018-06-10
Refactor and Improve Security

dshanske:
As part of this I,
* split the web sign in and the authorization code further. 
* rewrote the token generation code into a class to make management a bit easier and move some of that code out of the individual endpoint code
* added a second token generation class that uses transients to store the state.
* Cookie only stores a state property. Other parameters are stored as transients so they cannot be edited during the process. Future transients could store the IP address to confirm it was still part of the same transaction or other security measures.
* Rewrote the endpoint discovery code so it can discover any rel. Reason for this being future support for using it to discover 'redirect_uri' which is in the IndieAuth spec but not implemented. 
* Updated discovery unit tests to test new rel parsing.
* Tried to refactor along the lines of a possible future websignin plugin/indieauth plugin split.


includes/indieauth-functions.php
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	if ( 2 === (int) ( $code / 100 ) && isset( $response['me'] ) ) {
+		// The authorization endpoint acknowledged that the authorization code
+		// is valid and returned the authorization info
+		return $response;
+	}
+
+	// got an unexpected response from the authorization endpoint
+	$error = new WP_OAuth_Response( 'server_error', __( 'There was an error verifying the authorization code, the authorization server return an expected response', 'indieauth' ), 500 );
```
aaronpk: Message is wrong. Should be:

"the authorization server returned an unexpected response"

dshanske:
So, the latest updates further split the web signin and the endpoints. They now run independently inside the plugin. The settings page will be removed once the code is reviewed and commented on, as it is no longer needed. Websignin finds its endpoints based on the provided URL. The endpoints hard code their interactions and only work with each other.

pfefferle:
Can you please also fix the tests?

pfefferle:
I would like to keep them, because of the enable/disable web sign-in

dshanske:
I meant the Endpoint settings

dshanske:
I removed the test that isn't being used anymore. If you look, the warning is because the file is now empty. What I want to do is write some unit tests for the token classes I created to make sure that is working correctly, maybe test some of the individual functions as well. To protect against future bugs being introduced.

pfefferle:
Ok, but then please remove the file or add a dummy test, because I would love to have the tests green, before merging.

dshanske:
I will pull the endpoint settings tonight

dshanske:
And fix the tests

pfefferle:
@dshanske @aaronpk any last feedback (I changed some small things) or should I merge and release the new version?

miklb:
go forth and merge I say

aaronpk:
This looks great! Just testing it out now.

Looks like both IndieAuth login and authorization is working well, so glad to see the new local verification in there now. Also the web sign-in flow works great, I'm able to sign in to my test site using my aaronparecki.com identity (which is not running WordPress, also a good test).

Let's publish!

dshanske:
Yay!!!!

PR #74, opened 2018-06-27
Rewrite local links to SSL

dshanske:
I discovered this is a common problem for people creating issues.


pfefferle:
Is this a good idea? What happens if a site does not support https?

pfefferle (approved):

PR #81, opened 2018-07-12
Julfixes

dshanske:
Some minor tweaks. The biggest is that it tries to dispense with needing to map user to url by storing the user_id in the token. It falls back onto the me parameter if needed.


pfefferle (approved):
lgtm

PR #96, opened 2018-08-15
Switch to returning home_url as me over site_url

dshanske:
It has been pointed out this is inaccurate.

### DIFF
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -156,7 +156,7 @@ class IndieAuth_Authorization_Endpoint {
 			$this->delete_code( $code );
 
 			if ( class_exists( 'Indieweb_Plugin' ) && ( get_option( 'iw_single_author' ) || ! is_multi_author() ) ) {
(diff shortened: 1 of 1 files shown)


pfefferle (approved):

metbril:
I found a reference to site_url

https://github.com/indieweb/wordpress-indieauth/blob/fbf3a918f6469452193ba14a2aafd1ccb0bb4478/includes/functions.php#L200

Shouldn't that also be changed?

dshanske:
@metbril Would your site be on a different domain than your home?

metbril:
My site is hosted from root (`/`). You can browse, for example, an archive from https://robertvanbregt.nl/2018/

The software is installed in /wp. So the admin panel lives in https://robertvanbregt.nl/wp/wp-admin

(Although https://robertvanbregt.nl/admin will redirect just fine)

dshanske:
So this line you mentioned only checks for same domain

metbril:
I see. I'm still unsure if it needs updating. We'll wait and see. It's up to you. I only mentioned it because I ran into it when searching for the string.

PR #98, opened 2018-08-17
Improve Token Handling

dshanske:
This was prompted by the discovery that Omnibear won't work if you are logged into your site due cookie confllict. Still a problem, but I rearranged the logic to basically fail on a bad token, as opposed to passing control back to WordPress. With no token it will hand control to whatever came in, with a bad token it will hand back 0.

In order to test this, I made some minor modifications to the debug function.

Also, added a last accessed field so I could figure out which tokens I wanted to revoke. Testing generates a lot.


pfefferle (approved):
Code looks good to me, but I have not tested the functionality

PR #106, opened 2018-08-19
Error

dshanske:
Improve error handling to log to error_log if WP_DEBUG is on.

### DIFF
--- a/includes/class-oauth-response.php
+++ b/includes/class-oauth-response.php
@@ -26,7 +26,12 @@ class WP_OAuth_Response extends WP_REST_Response {
 	public function to_wp_error() {
 		$data   = $this->get_data();
 		$status = $this->get_status();
-		return new WP_Error( $data['error'], $data['error_description'], array( 'status' => $status, 'data' = $data ) );
+		return new WP_Error(
+			$data['error'], $data['error_description'], array(
+				'status' => $status,
+				'data'   => $data,
+			)
+		);
 	}
 
 	public function to_log() {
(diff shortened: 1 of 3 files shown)


includes/class-oauth-response.php
```diff
@@ -2,14 +2,20 @@
...
 		$this->set_data(
 			array(
 				'error'             => $error,
 				'error_description' => $error_description,
 			)
 		);
+		if ( is_array( $debug ) ) {
+			$this->set_debug( $debug );
+		}
+		if ( WP_DEBUG ) {
```
pfefferle: I would prefer to check if WP_DEBUG and WP_DEBUG_LOG are activated. If you encapsulate the logger into a global function, you do not have to check in each case, but only in the global logger function.

dshanske: Isn't using a log rather than the standard error log file a choice? I'm not sure everyone wants to use a special log file as opposed to the server log file. Also, this logs every error response created is WP_DEBUG is on.

dshanske: Therefore, I don't understand the request here. Can you elaborate? The error object has to be created any time there is an error response, regardless of whether it is being logged.

pfefferle: I do not mean to write a separate error log file... I vote for encapsulation the functionality in a global function like:

```PHP
function indieauth_error_log( $data ) {
  if ( WP_DEBUG && WP_DEBUG_LOG ) {
    error_log( $data );
  }
}
```
(pseudocode)

pfefferle: ah sorry, this is already in the request context, so it is already in a global context... misread that... forget my comment! But nevertheless, I would add the WP_DEBUG_LOG check, to only write to the log when it is explicitly enabled.

pfefferle (approved):

dshanske:
I write to the global error log in testing.

PR #108, opened 2018-08-23
Client Information Discovery

dshanske:
https://github.com/indieweb/indieauth/issues/23

Per this discussion, this PR is the intiial attempt to find info on the client_id page to enhance the token.


pfefferle (commented):
I would prefer the other way around: the setter/parser should select one single icon, the name, ... and the getter should only return the string.

dshanske:
@pfefferle Does this address your request? Everything is now private except direct retrieval.

pfefferle (approved):
👍

dshanske:
Ready for a release, you think?

pfefferle:
sure, release fast, release often!

PR #112, opened 2018-09-23
Add documentation and auth script

dshanske:
This updates the documentation to include some authorization issue diagnostics. It also bundles and links to @Zegnat 's authdiag script to check same. I didn't see the need to write a new one as it seems to do the trick.

### DIFF
--- a/authdiag.php
+++ b/authdiag.php
@@ -1,0 +1,126 @@
+<?php
+
+ini_set('display_errors', 1);
+error_reporting(E_ALL);
+
+$auth = filter_input(
+    INPUT_POST,
+    'auth',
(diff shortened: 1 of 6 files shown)


pfefferle (commented):
Some points:

* not WordPress coding standard complient.
* plugin files should not be directly accessible.
* should it be only accessible if "WP_DEBUG" is enabled?

dshanske:
I can make it cs compliant. Because this is for users who may not be as savvy, not sure about it being behind a debugger. Can try to reimplement without a direct call.

pfefferle (approved):
Feel free to merge it.

dshanske:
I don't see anything in the file that could be exploited.

PR #125, opened 2019-02-10
Febfixes

dshanske:
This plugin isn't showing as compatible with 5.0.3. Also, made a minor tweak to only show the headers on the front or author archive pages, as there is no point on other pages.


pfefferle (approved):
but I would release a minor change, because it changes the bahaviour how indieauth worked up to this change... even it it is only a small change

dshanske:
When you say minor change...are you suggesting this go to 3.2?

PR #127, opened 2019-02-24
Profile

dshanske:
Returns a jf2 h-card in the token return on verification so the client can display name and user avatar on an application

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -410,3 +410,29 @@ function indieauth_get_me() {
 	}
 	return $response['me'];
 }
+
+/* Returns jf2 formatted user data
+ *
+ */
+function indieauth_get_user( $user ) {
+	if ( is_numeric( $user ) ) {
+		$user = get_user_by( 'ID', $user );
+	}
+	if ( ! $user instanceof WP_User ) {
+		return array();
+	}
+	$return = array(
+		'type'  => 'card',
+		'name'  => $user->display_name,
+		'url'   => empty( $user->user_url ) ? get_author_posts_url( $user->ID ) : $user->user_url,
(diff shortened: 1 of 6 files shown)


pfefferle (approved):

includes/functions.php
```diff
@@ -410,3 +410,29 @@ function indieauth_get_me() {
...
+	if ( is_numeric( $user ) ) {
+		$user = get_user_by( 'ID', $user );
+	}
+	if ( ! $user instanceof WP_User ) {
+		return array();
+	}
+	$return = array(
+		'type'  => 'card',
+		'name'  => $user->display_name,
+		'url'   => empty( $user->user_url ) ? get_author_posts_url( $user->ID ) : $user->user_url,
```
pfefferle: shouldn't we always use the `author_posts_url`? Because the `user_url` might be a completely different domain.

pfefferle:
Is it required to send the correct `application/jf2feed+json` content-type?

https://www.w3.org/TR/jf2/#jf2feed_discovery

includes/functions.php
dshanske: Remember, this is a profile return. Not the 'me' return which is part of the client_id process, and shouldn't change. This is what the person wants to share as their site.

dshanske:
@pfefferle JSON needs to be the return for an Indieauth response.

pfefferle:
A, have to read more carefully... this is part of the response, not an extra endpoint protected by the access_token.

dshanske:
Theoretically, if someone wanted more than that, they could ask for it with a new scope, but this is just the basic info. It is now live in Quill, if you want to try logging into that after the update goes live.

Issue #128, opened 2019-03-09
Getting 403 errors On Some Applications


Changelingmx:
When trying to sign into several services, I get a 403 authorization error. These include but are not limited to Own Your Swarm, Indie Book Club, Monical, Own Your Checkin, and the indigenous app for iOS. When I run the test scrip on these, I get a message saying that the authorization method is depricated.

dshanske:
@Zegnat The authdiag script is yours. Looping you in.

Deprecated: filter_input(): explicit use of FILTER_FLAG_SCHEME_REQUIRED and FILTER_FLAG_HOST_REQUIRED is deprecated

Zegnat:
Those two specific flags have been [deprecated as of PHP 7.3](https://secure.php.net/manual/en/migration73.deprecated.php#migration73.deprecated.filter). I’ll see if I can sneak a PR in!

dshanske:
@Changelingmx I've merged in @Zegnat 's fix of the testing to be 7.3 compatible. However, not pushed to wordpress.org yet. That may not solve the problem though.

PR #129, opened 2019-03-10
Only explicitly specify URL requirements before PHP 7.3

Zegnat:
Super straight-forward fix for [PHP 7.3 deprecating the flags](https://secure.php.net/manual/en/migration73.deprecated.php#migration73.deprecated.filter) I was including. Addresses a comment in #128.

### DIFF
--- a/authdiag.php
+++ b/authdiag.php
@@ -101,19 +101,19 @@ if ( is_string( $url ) ) {
 		>
 	  <button type="submit">Run Test</button>
 	</form>
-<?php if ( $json !== null ) { ?>
+<?php if ( null !== $json ) { ?>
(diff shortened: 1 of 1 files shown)


Zegnat:
@pfefferle does this repository include a command to test against the standard?

Zegnat:
Got it to run using `php vendor/squizlabs/php_codesniffer/bin/phpcs authdiag.php`. Fixed everything that is fixable, please review the commits.

Note that this file may be [considered as being an external resource](https://chat.indieweb.org/wordpress/2019-03-10/1552226894251100) rather than code of the plugin, so the style is less important.

PR #130, opened 2019-03-11
PKCE Support

dshanske:
This is a first pass at adding PKCE Support. Also switches all code and token generate to SHA256 from the default WordPress salted MD5. It also bumps the minimum PHP version to 5.4.

Looking to see if this appears to be implemented correctly as this is my first time trying this.


dshanske:
This definitely needs more work and testing. I can't get it to fail.

aaronpk (approved):
I tested this out with my test client and threw a variety of inputs at it, successful and error conditions. Looks like this is doing what it's supposed to be doing! Looks great! :tada:

PR #138, opened 2019-05-05
Add Bulk Actions and Improve List/Token Generation

dshanske:
This improves the token list table, adding the ability to revoke tokens based on last accessed dates, as well as cleaning up expired tokens and blowing out authorization codes if there. It also adds the expiration to the list table for future expiring tokens as limited support is there.

Also, the scopes for the table generated scopes were not configurable and now they are.


dshanske:
@pfefferle If you could look at this one, I'll release 3.3.2

dshanske:
@pfefferle Added a notation re 5.2 compatibility.

includes/class-token-list-table.php
```diff
@@ -32,37 +46,95 @@ public function prepare_items() {
...
 		}
+
 	}
 
 	public function column_default( $item, $column_name ) {
 		return $item[ $column_name ];
 	}
 
+	public function column_cb( $item ) {
+		return sprintf( '<input type="checkbox" name="tokens[]" value="%s" />', $item['token'] );
```
pfefferle: I would recommend to output-escape the token.

includes/class-token-list-table.php
```diff
@@ -32,37 +46,95 @@ public function prepare_items() {
...
+	}
+
 	public function process_action() {
-		$tokens = isset( $_GET['tokens'] ) ? $_GET['tokens'] : array(); // phpcs:ignore
+		$tokens = isset( $_REQUEST['tokens'] ) ? $_REQUEST['tokens'] : array(); // phpcs:ignore
```
pfefferle: I would recommend to sanatize the request params.

includes/class-token-list-table.php
```diff
@@ -7,15 +7,29 @@
...
 
+	public function get_bulk_actions() {
+			  return array(
+				  'revoke'   => __( 'Revoke', 'indieauth' ),
+				  'revokey'  => __( 'Revoke Tokens Last Accessed 1 Year Ago', 'indieauth' ),
+				  'revokem'  => __( 'Revoke Tokens Last Accessed 1 Month Ago', 'indieauth' ),
+				  'revokew'  => __( 'Revoke Tokens Last Accessed 1 Week Ago', 'indieauth' ),
+				  'revoked'  => __( 'Revoke Tokens Last Accessed 1 Day Ago', 'indieauth' ),
+				  'revokeh'  => __( 'Revoke Tokens Last Accessed 1 Hour Ago', 'indieauth' ),
+				  'maintain' => __( 'Clean Up Expired Tokens and Authorization Codes', 'indieauth' ),
```
pfefferle: the keys are not very descriptive, especially "maintain".

pfefferle: I would prever at least `revoke_*` and `cleanup` instead of `maintain`.

pfefferle (approved):
I have only reviewed the code, I have not tested the functionality.

includes/class-indieauth-token-ui.php
```diff
@@ -54,7 +54,12 @@ public function new_token() {
 			exit;
 		}
 		require ABSPATH . 'wp-admin/admin-header.php';
-		$token = self::generate_local_token( $_REQUEST['client_name'] );
+		$scopes = trim( implode( ' ', $_REQUEST['scopes'] ) );
```
pfefferle: sanatize?

PR #148, opened 2019-07-07
Fixes

dshanske:
Discovered that the code as previously written was not associating tokens properly based on cnditionals for the Indieweb plugin. This should fix that and it is locked to either returning the root URL or the author URL.

Also, for web-signin, to ensure security, unique URL properties for user accounts are now enforced. URLs can now be seen when viewing all users.

Also added IndieAuth to the site health check and changed the language in the success/failure scripting. 

Need to do some additional scope hardening, but this should fix the problem I discovered this week until I can.


includes/class-indieauth-admin.php
```diff
@@ -11,6 +11,96 @@ public function __construct() {
...
+		}
+		return $val;
+	}
+
+	public function add_user_url_column( $column ) {
+		$column['user_url'] = __( 'Website', 'indieauth' );
+		return $column;
+	}
+
+			/**
```
pfefferle: indents!

includes/class-indieauth-admin.php
```diff
@@ -62,9 +151,9 @@ public function admin_init() {
 		add_settings_field( 'indieauth_general_settings', __( 'IndieAuth Settings', 'indieauth' ), array( $this, 'general_settings' ), 'general', 'default' );
 	}
 
-	/**
-	 * Add IndieAuth options to the WordPress general settings page.
-	 */
+			/**
```
pfefferle: indents!

includes/class-indieauth-admin.php
```diff
@@ -75,9 +164,9 @@ public function general_settings() {
 		printf( __( 'Based on your feedback and to improve the user experience, we decided to move the settings to a separate <a href="%1$s">settings-page</a>.', 'indieauth' ), $path ); // phpcs:ignore
 	}
 
-	/**
-	 * Add admin menu entry
-	 */
+			/**
```
pfefferle: indents!

includes/class-indieauth-admin.php
```diff
@@ -123,23 +209,35 @@ public function settings_page() {
...
 			$json = json_decode( wp_remote_retrieve_body( $response ) );
-			set_query_var( 'authdiag_message', $json->message );
+			return $json->message;
 		} else {
-			set_query_var( 'authdiag_message', 'Fail' );
+			return false;
 		}
+	}
 
+			/**
```
pfefferle: indents!

indieauth.php
```diff
@@ -3,7 +3,7 @@
  * Plugin Name: IndieAuth
  * Plugin URI: https://github.com/indieweb/wordpress-indieauth/
  * Description: IndieAuth is a way to allow users to use their own domain to sign into other websites and services
- * Version: 3.3.2
+ * Version: 3.3.3
```
pfefferle: This sounds greater than a "patch"?

readme.txt
```diff
@@ -125,6 +125,10 @@ Some hosting providers filter this out using mod_security. For one user, they ne
 
 == Upgrade Notice ==
 
+= 3.3.3 =
+
+Due to the possibility of someone setting the url in their user profile to the same as another account, you will no longer be able to save the exact same url into two accounts
```
pfefferle: what if two already use the same domain?

indieauth.php
dshanske: You think it should bump to 3.4? It doesn't seem to bring any new features...although I suppose the unique URL thing is a feature.

readme.txt
dshanske: The system would discard the second one and give anyone access to the first. But, you think it should be flagged and addressed?

dshanske:
I will fix the indents. The biggest reason for these changes was the possibility that the system as it existed would return a token for a user other than the one who logged in.

indieauth.php
pfefferle: and it breaks backwards compatibility... if only a little ;)

dshanske: Issuing corrections now.

dshanske:
@pfefferle Bumped to 3.4.0

pfefferle (commented):
Shouldn't it throw an error if two users use the same URL? Otherwise it still is a security issue!

readme.txt
pfefferle: Shouldn't it throw an error if two users use the same URL? Otherwise it still is a security issue!

dshanske:
Right now, it just silently refuses to save. I have to figure out how to generate an error on saving. I don't see anything in the filter to allow that.

pfefferle:
Not while saving... When someone tries to login with an URL, that another person also uses, you said, the first one is used, but it should throw an error! Something like "Not possible to identify unique urser" or something like that... Otherwise it does not fix the issue. At least not, until all have changed their URLs...

dshanske:
I just can't figure out how to throw that error. I don't see a way to surface it. I may have to dig deeper into the code.

pfefferle:
Instead of the "Approve" page?

dshanske:
You mean, check during the process the plugin controls? Interesting. That could work.

dshanske:
@pfefferle This adds a few things. If the count of returns in searching for a URL is more than 1 it will now fail. And it does a check on the user list page for dupes.

dshanske:
@pfefferle This ready to go?

pfefferle (approved):

PR #153, opened 2019-08-03
Add setting for site URL user

dshanske:
The recent changes to the plugin exposed a problem with using the Indieweb plugin's single author setting. So this plugin now has a setting for deciding who gets to use the site's URL, removing the dependency. It still, if no value is set, wil use the Indieweb plugin value or if there is only one user, that user. But this is much more reliable.

### DIFF
--- a/includes/class-indieauth-admin.php
+++ b/includes/class-indieauth-admin.php
@@ -170,6 +170,16 @@ class IndieAuth_Admin {
 				'default'      => 0,
 			)
 		);
+		register_setting(
+			'indieauth',
+			'indieauth_root_user',
+			array(
+				'type'         => 'int',
+				'description'  => __( 'User Who is Represented by the Site URL', 'indieauth' ),
+				'show_in_rest' => true,
+				'default'      => get_option( 'iw_default_author', 0 ),
+			)
+		);
 	}
 
 	public function admin_init() {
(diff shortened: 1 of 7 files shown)


dshanske:
@pfefferle Want to see if I can get this out. I had a comment from someone that the last update was causing their IndieAuth URL to change to /author/username instead of root.

The changes in the last version made things more secure on that, but this part was still fragile.

includes/class-indieauth-admin.php
```diff
@@ -170,6 +170,16 @@ public function settings() {
...
 			)
 		);
+		register_setting(
+			'indieauth',
+			'indieauth_root_user',
+			array(
+				'type'         => 'int',
+				'description'  => __( 'User Who is Represented by the Site URL', 'indieauth' ),
+				'show_in_rest' => true,
+				'default'      => get_option( 'iw_default_author', 0 ),
```
pfefferle: what if there is no user with the id `0`?

pfefferle: Seems to be a similar discussion we had in the #indieweb-wordpress chat.

pfefferle: or is it showing `None` in this case?

pfefferle:
Is the fallback to "the first user in the db" really a good idea? https://github.com/indieweb/wordpress-indieauth/pull/153/files#diff-f81f3dff36bf04b7a070d642c858e883R190

Shouldn't we force to use the settings if there is no "single author"?

includes/class-indieauth-admin.php
pfefferle: can we use the same mechanism in the indieweb plugin?

pfefferle (approved):

includes/class-indieauth-admin.php
dshanske: It's showing none. Then it goes back to using author urls or checking for a single user.

PR #167, opened 2020-01-25
Restore Remote Capability

dshanske:
This adds an HTTP/SSL check to the Site Health system, restores the remote endpoint functionality, and introduces loading logic to either allow the site to run using an external service like IndieAuth or using the built-in service without loading both simultaneously. It also turns IndieAuth_Plugin into a static class. We can't remove the class as Micropub and Microsub at this time check for its presence. The purpose of the class is to handle the load and store the object reflecting the remote/local configuration.

I tried to get the language in the settings clear about what each does.

### DIFF
--- a/includes/class-indieauth-remote-authorize.php
+++ b/includes/class-indieauth-remote-authorize.php
@@ -1,0 +1,168 @@
+<?php
+
+/**
+ * Micropub IndieAuth Authorization Class
+ */
+class IndieAuth_Remote_Authorize extends IndieAuth_Authorize {
+
+
+	public function __construct( $load = true ) {
+		$this->register_settings();
+		// Load the hooks for this class only if true. This allows for debugging of the functions
+		if ( ! $load ) {
+			return;
+		}
+		add_action( 'admin_init', array( get_called_class(), 'admin_init' ) );
+		$this->load();
+	}
+
+	public function register_settings() {
+		// Register Setting
+		register_setting(
+			'indieauth',
+			'indieauth_authorization_endpoint', // Setting Name
+			array(
+				'type'              => 'string',
+				'description'       => 'IndieAuth Authorization Endpoint',
+				'sanitize_callback' => 'esc_url',
+				'show_in_rest'      => true,
+			)
+		);
+		// Register Setting
+		register_setting(
+			'indieauth',
+			'indieauth_token_endpoint', // Setting Name
+			array(
+				'type'              => 'string',
+				'description'       => 'IndieAuth Token Endpoint',
+				'sanitize_callback' => 'esc_url',
+				'show_in_rest'      => true,
+			)
+		);
+
+	}
+
+	public static function admin_init() {
+		$cls  = get_called_class();
+		$page = 'indieauth';
+				add_settings_section(
+					'indieauth',
+					'Remote IndieAuth Endpoint Settings',
+					array( $cls, 'auth_settings' ),
+					$page
+				);
+				add_settings_field(
+					'indieauth_authorization_endpoint',
+					__( 'Authorization Endpoint', 'indieauth' ),
+					array( $cls, 'endpoint_field' ),
+					$page,
+					'indieauth',
+					array(
+						'label_for' => 'indieauth_authorization_endpoint',
+						'class'     => 'widefat',
+						'default'   => 'https://indieauth.com/auth',
+					)
+				);
+				add_settings_field(
+					'indieauth_token_endpoint',
+					__( 'Token Endpoint', 'indieauth' ),
+					array( $cls, 'endpoint_field' ),
+					$page,
+					'indieauth',
+					array(
+						'label_for' => 'indieauth_token_endpoint',
+						'class'     => 'widefat',
+						'default'   => 'https://tokens.indieauth.com/token',
+					)
+				);
+	}
+
+	public static function endpoint_field( $args ) {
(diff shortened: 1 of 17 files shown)


pfefferle:
Indents! 😉

dshanske:
@pfefferle I will go through and look for incorrect indents. How is it otherwise?

dshanske:
@pfefferle Was that the right indent?

includes/class-indieauth-remote-authorize.php
```diff
@@ -0,0 +1,180 @@
...
+				'client_id'    => home_url(),
+			)
+		);
+		$resp  = wp_safe_remote_post( static::get_authorization_endpoint() . '?' . $query, $args );
+		if ( is_wp_error( $resp ) ) {
+			return wp_error_to_oauth_response( $resp );
+		}
+		$code   = wp_remote_retrieve_response_code( $resp );
+		$body   = wp_remote_retrieve_body( $resp );
+		$params = json_decode( $response, true );
```
pfefferle: `$params` is never used?

dshanske: I see the mistake. Fixing now.

dshanske: After I tested it, I noticed one function used params, the other reused body. I matched them, but made a mistake. 

I have more PRs before I propose releasing a new version. I want to do something with scopes. I also want some feedback on the settings text.

dshanske:
@pfefferle Is this ready to merge?

pfefferle:
I had not yet the time to look at the code, sorry!

dshanske:
@pfefferle I thought you had, misunderstood, sorry

dshanske:
Rebased this based on the merge of the other PR

dshanske:
@pfefferle Any chance to look at this one?

readme.txt
```diff
@@ -3,8 +3,8 @@ Contributors: indieweb, pfefferle, dshanske
 Tags: IndieAuth, IndieWeb, IndieWebCamp, login
 Requires at least: 4.9.9
 Requires PHP: 5.4
-Tested up to: 5.2.2
-Stable tag: 3.4.2
+Tested up to: 5.3.2
```
pfefferle, line 6: should 5.4 now

dshanske: I can fix that. How's everything else?

pfefferle:
The settings are a bit complicated. You choose "Use Another Site as an IndieAuth Endpoint" and you have to hit save before you see the settings for the custom endpoints, you have to change them and then hit enter again.

I would prefer to add the fields below the "Use Another Site as an IndieAuth Endpoint" option, so the user is able to edit them in the same step as he changes the option.

dshanske:
> The settings are a bit complicated. You choose "Use Another Site as an IndieAuth Endpoint" and you have to hit save before you see the settings for the custom endpoints, you have to change them and then hit enter again.
> 
> I would prefer to add the fields below the "Use Another Site as an IndieAuth Endpoint" option, so the user is able to edit them in the same step as he changes the option.

That works. I want to see if I  can get some feedback on the UI part as well.

pfefferle:
Hide "In order to ensure IndieAuth tokens will work please visit the settings page to check: Visit Settings Page" error on the IndieAuth settings page.

pfefferle:
Show "Authorization has Failed" description directly in the site-health check?

dshanske:
> Show "Authorization has Failed" description directly in the site-health check?

I thought I did that?

dshanske:
> Hide "In order to ensure IndieAuth tokens will work please visit the settings page to check: Visit Settings Page" error on the IndieAuth settings page.

That I will change. Might be to run a health check.

pfefferle:
> > Show "Authorization has Failed" description directly in the site-health check?

> I thought I did that?

Only the reference to the settings site with the big red header... but why not show the "howto fix that problem" directly to the health page? So we might be able to remove the big red box from the settings page some time in the future 😉

dshanske:
@pfefferle I moved all the error checking into the Site Health functionality. I was looking at the other question, re the fields to edit the endpoints...if you notice, those settings are not loaded until you load the remote class. I suppose I could register them regardless, then use a little JS to unhide them on the page.

tw2113:
I see nothing standing out as needing addressed at this point. Hesitant to approve just yet because I think @pfefferle may have a bit more first.

pfefferle:
> if you notice, those settings are not loaded until you load the remote class. I suppose I could register them regardless, then use a little JS to unhide them on the page.

I can comprehend your design decisions, but it is a new old core functionality of the plugin, so why not move the settings stuff to the settings classes/templates and only "lazy" load the functionality?

I also do not see the need of JS. I would simply load the form items under the second check box and ignore them if the first checkbox is selected.

Bildschirmfoto 2020-04-21 um 06 45 15

BTW. the code renders two labels with the same `for`:

```html
<th scope="row">
  <label for="indieauth_authorization_endpoint">Authorization Endpoint</label>
</th>
<td>
  <label for="indieauth_authorization_endpoint">
    <input id="indieauth_authorization_endpoint" name="indieauth_authorization_endpoint" type="url" value="https://indieauth.com/auth" placeholder="https://indieauth.com/auth">
  </label>
</td>
```

dshanske:
@pfefferle I'd like to merge this PR. I've removed the selector for the remote endpoint for now to allow the other enhancements and refactoring to be merged. See https://chat.indieweb.org/dev/2020-07-14/1594761683886700 

There are still people who want to not host their own endpoint and delegate it, but for now, want to make this an 'expert' feature.

dshanske:
The most recent change keeps the libraries for remote code and the enhancements to allow it in the plugin. It removes the configuration option that will enable them. This will, in a subsequent PR, become an 'expert mode' that will not be visible to users by default. Indieauth.com is no longer mentioned or used anywhere in the code. 

There are several needed steps before release related to this that will be pushed in future PRs, including #168 .

PR #169, opened 2020-02-01
Implement Scope Support

dshanske:
This starts the process of adding scope support to the plugin. What it does is create a class that defines a scope as a collection of WordPress capabilities.

It then defines a Scopes object that stores the scope definitions and adds a filter to map_meta_cap which, if there is an indieauth process going on, it will compare the list of capabilities required to the ones in the scope, and if any of the requires ones are missing, it will revoke permission.

This means that any plugin can just use current_user_can instead of managing scopes, as they currently have to do. 

There is an issue in that Microsub does not have any specific capabilities built into WordPress. Will be looking into that in the future, but until then, full permissions are granted to all users, as the read capability is used by all. This would require some custom capabilities that could be added to roles. 

This is instantiated within the file, but if merged, I will change that in the PR I agreed to do after this and the other pending PR are merged.


dshanske:
Thinking of changing the meta cap filter to fire later. Thoughts?

includes/class-indieauth-scopes.php
```diff
@@ -0,0 +1,222 @@
...
+				)
+			)
+		);
+
+		$this->register_scope(
+			new IndieAuth_Scope(
+				'delete',
+				__( 'Allows the application to delete or undelete posts', 'indieauth' ),
+				array(
+					'delete_posts',
```
pfefferle: doesn't this also require `edit_posts` or is this implicit?

includes/class-indieauth-scopes.php
```diff
@@ -0,0 +1,222 @@
...
+				__( 'Allows the application read access to Microsub channels', 'indieauth' ),
+				array(
+					'read',
+				)
+			)
+		);
+
+		$this->register_scope(
+			new IndieAuth_Scope(
+				'follow',
```
pfefferle: Shouldn't this be added by the Microsub plugin? Shouldn't the IndieAuth plugin as generic as possible?

pfefferle (approved):
I had only a look at the code and din't test the functionality.

(perhaps you could also some more php-doc-blocks)

pfefferle:
Nice idea btw.!

includes/class-indieauth-scopes.php
dshanske: edit_posts would be the permission for update

includes/class-indieauth-scopes.php
dshanske: It is...these are all the known scopes for IndieAuth right now.

pfefferle: But the comment is very specific with Microsub.

dshanske:
Working on adding a few docblocks, then I will merge and rebase the other pr

dshanske:
Adding some unit tests as well, before merge, just to check for issues.

PR #186, opened 2020-10-04
Tries to Adopt some of the New IndieAuth Changes

dshanske:
The most notable changes here are cuts. The unique URL code is gone, as per the changes to the spec, only URLs on the same domain as the initially entered URL should be accepted by the client. 

The me url is now optional, so changes were added to ensure less reliance on it. The URL returned is either the root domain when set in the plugin, or the author URL. This ensures, as per the concerns in https://indieauth.spec.indieweb.org/#differing-user-profile-urls.

This also changes the profile response, which was experimental, to the final version, which only appears with a profile scope being requested, and adds support for the email scope.

Since the exchange has been simplified, the 'id' response_type is now gone, and the new parameters for issuing a token vs just a profile return have been added. Because not all clients may have caught up, until further notice, requesting 'id' will behave like a 'code' request with no scope.

### DIFF
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -80,7 +80,9 @@ class IndieAuth_Authorization_Endpoint {
 						/* grant_type=authorization_code is the only one supported right now. This remains optional as not required in
 						 * the original IndieAuth spec, but will eventually be mandatory.
 						 */
-						'grant_type'    => array(),
+						'grant_type'    => array(
+							'default' => 'authorization_code',
+						),
 						/* The authorization code received from the authorization endpoint in the redirect.
 						 */
 						'code'          => array(),
@@ -197,12 +199,15 @@ class IndieAuth_Authorization_Endpoint {
 	public function verify( $request ) {
 		$params = $request->get_params();
 
-		$required = array( 'redirect_uri', 'client_id', 'code' );
+		$required = array( 'redirect_uri', 'client_id', 'code', 'grant_type' );
 		foreach ( $required as $require ) {
 			if ( ! isset( $params[ $require ] ) ) {
 				// translators: Name of missing parameter
 				return new WP_OAuth_Response( 'parameter_absent', sprintf( __( 'Missing Parameter: %1$s', 'indieauth' ), $require ), 400 );
 			}
+		}
+		if ( 'authorization_code' !== $params['grant_type'] ) {
+			return new WP_OAuth_Response( 'invalid_grant', __( 'Endpoint only accepts authorization_code grant_type', 'indieauth' ), 400 );
 		}
 		$params = wp_array_slice_assoc( $params, array( 'client_id', 'redirect_uri' ) );
 		$code   = $request->get_param( 'code' );
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -29,42 +29,44 @@ class IndieAuth_Authorization_Endpoint {
 						/* Code is currently the only type as of IndieAuth 1.1 and a response_type is now required, but not requiring it here yet.
 						 * Indicates to the authorization server that an authorization code should be returned as the response.
 						 */
-						'response_type' => array(
-							'default'           => 'code',
-							),
+						'response_type'         => array(
+							'default' => 'code',
+						),
 						// The Client URL.
-						'client_id'     => array(
+						'client_id'             => array(
 							'validate_callback' => 'rest_is_valid_url',
 							'sanitize_callback' => 'esc_url_raw',
+							'required'          => true,
 						),
 						// The redirect URL indicating where the user should be redirected to after approving the request.
-						'redirect_uri'  => array(
+						'redirect_uri'          => array(
 							'validate_callback' => 'rest_is_valid_url',
 							'sanitize_callback' => 'esc_url_raw',
+							'required'          => true,
 						),
-						/* A parameter set by the client which will be included when the user is redirected back to the client. 
+						/* A parameter set by the client which will be included when the user is redirected back to the client.
 						 * This is used to prevent CSRF attacks. The authorization server MUST return the unmodified state value back to the client.
 						 */
-						'state'         => array(
-							'required'          => true
+						'state'                 => array(
+							'required' => true,
 						),
 						/* Code Challenge.
 						 * IndieAuth 1.1 requires PKCE, but for now these parameters will remain optional to give time for other implementers.
 						 */
-						'code_challenge' => array(),
+						'code_challenge'        => array(),
 						/* The hashing method used to calculate the code challenge, e.g. "S256"
 						 */
 						'code_challenge_method' => array(),
 
-						/* A space-separated list of scopes the client is requesting, e.g. "profile", or "profile create". 
-						 * If the client omits this value, the authorization server MUST NOT issue an access token for this authorization code. 
-						 * Only the user's profile URL may be returned without any scope requested. See Profile Information for details about 
+						/* A space-separated list of scopes the client is requesting, e.g. "profile", or "profile create".
+						 * If the client omits this value, the authorization server MUST NOT issue an access token for this authorization code.
+						 * Only the user's profile URL may be returned without any scope requested. See Profile Information for details about
 						 * which scopes to request to return user profile information. Optional.
 						 */
-						'scope' => array(),
+						'scope'                 => array(),
 						/* The Profile URL the user entered. Optional.
 						 */
-						'me'            => array(
+						'me'                    => array(
 							'validate_callback' => 'rest_is_valid_url',
 							'sanitize_callback' => 'esc_url_raw',
 						),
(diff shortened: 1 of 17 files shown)


pfefferle:
One test is failing.

dshanske:
> One test is failing.

Sorry about that. Forgot to remove a test that is now deprecated.

dshanske:
@pfefferle Can I ask you to check again?

includes/class-indieauth-authorization-endpoint.php
```diff
@@ -238,34 +287,46 @@ public function authorize() {
 			)
 		);
 		$url    = add_query_params_to_url( $args, wp_login_url() );
-		if ( 'code' === $_GET['response_type'] ) {
-			include plugin_dir_path( __DIR__ ) . 'templates/indieauth-authorize-form.php';
-		} elseif ( 'id' === $_GET['response_type'] ) {
+		if ( empty( $scopes ) || empty( array_diff( $scopes, array( 'profile', 'email' ) ) ) || array( 'profile' ) === $scopes ) {
```
Zegnat:
```suggestion
		if ( empty( $scopes ) || empty( array_diff( $scopes, array( 'profile', 'email' ) ) ) ) {
```

I think the last `||` can be removed? The `empty` check before will already return true if `$scopes` only contains “profile”.

includes/class-indieauth-authorization-endpoint.php
```diff
@@ -238,34 +287,46 @@ public function authorize() {
...
 		$current_user = wp_get_current_user();
+		$user         = $current_user->ID;
 		// phpcs:disable
 		$client_id     = wp_unslash( $_POST['client_id'] ); // WPCS: CSRF OK
 		$redirect_uri  = isset( $_POST['redirect_uri'] ) ? wp_unslash( $_POST['redirect_uri'] ) : null;
 		$scope         = isset( $_POST['scope'] ) ? $_POST['scope'] : array();
 		$code_challenge  = isset( $_POST['code_challenge'] ) ? wp_unslash( $_POST['code_challenge'] ) : null;
 		$code_challenge_method  = isset( $_POST['code_challenge_method'] ) ? wp_unslash( $_POST['code_challenge_method'] ) : null;
+
+		// Do not allow the post scope as deprecated. For compatibility, instead update the offering to the more limited but functionally identical create/update.
```
Zegnat, line 327: Does the code here never allow me to request a “post” scope? Why? Is this a WordPress-only thing?

includes/class-indieauth-token-endpoint.php
```diff
@@ -151,84 +161,64 @@ public function request( $params ) {
...
+			// Return only the standard keys in the response
+			return new WP_REST_Response(
+				wp_array_slice_assoc(
+					$return,
+					array(
+						'access_token',
+						'token_type',
+						'scope',
+						'me',
+						'profile',
```
Zegnat, line 221: Maybe I am missing something, but when is `profile` added to `$return`? Line 182 adds it to `$token`. And with the removal of the old line 227, I am not sure when it is being added at all.

Zegnat (commented):
Seems to be a lot of WP things that I can’t really comment on. But just from going through the logic in the PR, here are some quick notes.

aaronpk:
While you're at it, you should drop the requirement of the `me` parameter in the token request. 

https://indieauth.spec.indieweb.org/#request

includes/class-indieauth-authorization-endpoint.php
dshanske: Because post is deprecated. So it is being updated to create/update.

includes/class-indieauth-token-endpoint.php
dshanske: Fixing that one.

includes/class-indieauth-authorization-endpoint.php
Zegnat: Nitpick: IndieAuth does not deprecate scopes.

I guess this is because Micropub is using something else now and you want to be helpful there? Only nitpicking because something else may want to use IndieAuth as an easy OAuth provider and still want access to post scopes for their purposes. Not really important, just separation of concerns thing.

dshanske: I didn't deprecate it, it just turns it into create update so Micropub doesn't have to support a more open permission.

dshanske:
@pfefferle Would you be able to review and approve this? I updated to match the November spec, did several feature checks. I also adopted storing a UUIID and Last IP address in the token to match the Application Password core implementation, but will likely file a separate PR for a few other ideas I have based on their code.

includes/class-indieauth-authorization-endpoint.php
```diff
@@ -218,7 +272,7 @@ public function authorize() {
 		$client_icon = $info->get_icon();
 		$redirect_uri  = isset( $_GET['redirect_to'] ) ? wp_unslash( $_GET['redirect_to'] ) : null;
 		$scope         = isset( $_GET['scope'] ) ? wp_unslash( $_GET['scope'] ) : null;
-		$scopes        = explode( ' ', $scope );
+		$scopes        = array_filter( explode( ' ', $scope ) );
 		$state         = isset( $_GET['state'] ) ? $_GET['state'] : null;
 		$me            = isset( $_GET['me'] ) ? wp_unslash( $_GET['me'] ) : null;
```
pfefferle, line 290: I know that this is old code, but shouldn't the `me` param be checked for the same domain? It seems to be no security isse, but it looks strange if I change the `me` get param on the `autorisation` page.

See:

Bildschirmfoto 2021-01-05 um 17 04 06

dshanske: Actually, it should remove me entirely and instead use the user's URL. If you check the spec changes, the me sent in is a hint. So, since we are always using the logged in user, it should be replaced with either the root domain or /author/username as outlined elsewhere in the code

dshanske:
@pfefferle Thanks for looking at this. I'm excited to get it out.

includes/class-indieauth-authorization-endpoint.php
dshanske: There was never any danger of spoofing another URL. But the new version won't allow you to spoof the displayed version.

includes/class-indieauth-local-authorize.php
```diff
@@ -35,6 +35,7 @@ public function verify_access_token( $token ) {
 			return $return;
 		}
 		$return['last_accessed'] = time();
+		$return['last_ip']       = $_SERVER['REMOTE_ADDR'];
```
Ruxton, line 38: Shouldn't this be a little more robust?  and do something like..

```php
function getUserIpAddr(){
  if(!empty($_SERVER['HTTP_CLIENT_IP'])) {
    $ip = $_SERVER['HTTP_CLIENT_IP'];
  }  elseif(!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
    $ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
  } else{
    $ip = $_SERVER['REMOTE_ADDR'];
  }
  return $ip;
}
```

templates/indieauth-authenticate-form.php
```diff
@@ -13,7 +13,8 @@
 			printf(
 				'<p>' . __( 'The app <strong>%1$s</strong> would like to sign you in as <strong>%2$s</strong>.', 'indieauth' ) . '</p>',
 				$client_id,
-				$me
+				get_url_from_user( $current_user->ID )
```
Ruxton: there's a very very small possibility of the return null value coming through and it's not being handled here.  There should be a better textual representation of null in this situation.

readme.md
```diff
@@ -130,6 +130,11 @@ Some hosting providers filter this out using mod_security. For one user, they ne
 
 ## Upgrade Notice ##
 
+### 3.6.0 ###
```
Ruxton, line 133: Having this is in the README is overly noisey. Upgrade Notice and Changelog are only going to grow and grow and grow.  Could we shift them out to UPGRADE and CHANGELOG respectively?

templates/indieauth-authorize-form.php
```diff
@@ -33,8 +33,10 @@
 		<?php _e( 'The app is requesting the following <a href="https://indieweb.org/scope">scopes</a>', 'indieauth' ); ?>
 		<ul>
 		<?php
-		foreach ( $scopes as $s ) {
-			printf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), self::scopes( $s ) );
+		if ( ! empty( $scopes ) ) {
```
Ruxton: you should make a func for this that accepts $scopes.

The code at @https://github.com/indieweb/wordpress-indieauth/pull/186/files#diff-d55842c372a35c31dfc5037ff0ffbc250d02cda9da20b6f2edcc4502d0429f60R31-R35 is exactly the same, so pull it out and func it up.

Ruxton (changes requested):
Minor changes and feedback, otherwise looks good.

readme.md
pfefferle: The `README.md` is generated from the `readme.txt`, so that's why it is part of the readme. I would not change that in the first place, because then we have to write a tool or do a lot by hand.

templates/indieauth-authenticate-form.php
pfefferle: Is there really a chance? If the user is not logged in, he never see this site.

includes/class-indieauth-local-authorize.php
dshanske: I added it to match the fact that the new core Application Passwords code stores this. I was trying to bring this implementation closer to that for future coding considerations. Same reason I added a uuid. It only uses REMOTE_ADDR. As does the comment implementation that stores IP.

https://github.com/WordPress/WordPress/blob/1ead0719b7897632d4d44309dbd904d9eb96dd28/wp-includes/class-wp-application-passwords.php#L238

readme.md
dshanske: The various sections of the file are used by WordPress Core in the readme.txt file, so they need to be there. I suppose if we wanted to, we could put the UPGRADE, CHANGELOG, etc into separate files, and add a simple script that cats them together for the readme.txt but not for the readme.md. But...back to tooling.

templates/indieauth-authenticate-form.php
Ruxton: it takes a minute to defend against it. imo the fact the function is capable of returning null means it should get handled. I don't think you can guarantee the state of current_user, as other plugins CAN mess it up.

dshanske:
Didn't think those issues were blockers, but resolved them.

readme.md
dshanske: Would suggest, being as this is a unrelated to the PR at hand, we consider this separately from this revision.

Ruxton (approved):

pfefferle (approved):
currently it is required that someone with "write access" has to review... I just changed that requirement, so it should be enough to have a `+1` review for the next review!

pfefferle:
Thanks for your review @Ruxton !

PR #191, opened 2021-02-16
Update consent screens

dshanske:
This is all trying to update the consent screens. In light of recent discussions in chat, I thought they needed a refresh on the display and language.  This updates both the authorize and the authenticate screens.


pfefferle (changes requested):
I like the changes, but `&#9919;` does not show properly on my Mac with Firefox. Can we use https://emojipedia.org/locked/ instead?

dshanske:
That's interesting...I use Firefox on Linux. I thought an HTML code would be more universal. I'll change it

dshanske:
@pfefferle Done.

pfefferle (approved):

PR #194, opened 2021-02-20
Clean Up Token Job

dshanske:
This introduces the indieauth_cleanup schedule, which is run twice daily and cleans up any expired authorization codes or tokens. The cron job is activated on plugin update or activation or if not scheduled when visiting the token page, and removed on deactivation.

To accommodate this, as the token class wasn't designed for multiple users, added a function to find all users who have tokens, and, as a LIKE query can use up resources, the results are saved in the options table, and invalidated any time the user setting a token is not on the list. The list is also refreshed when you visit the token table. 

The function to retrieve all tokens for a user now checks expiry when it does, and will delete any expired tokens at that point as well. As this is now scheduled, the manual cleanup option was removed.

This is a precursor to adding an option for expiring tokens. Technically, the code already supported expiring tokens, but not actually setting a token to expire. The token retrieval code checks the expiry date and destroys the token if it is expired. This runs that check on a schedule, which means tokens that are expired will not wait until someone tries to use them to be deleted.

### DIFF
--- a/includes/class-token-user.php
+++ b/includes/class-token-user.php
@@ -101,8 +101,13 @@ class Token_User extends Token_Generic {
 		foreach ( $meta as $key => $value ) {
 			if ( 0 === strncmp( $key, $this->prefix, strlen( $this->prefix ) ) ) {
 				$value         = maybe_unserialize( array_pop( $value ) );
+				$key = str_replace( $this->prefix, '', $key );
 				$value['user'] = $this->user_id;
-				$tokens[ str_replace( $this->prefix, '', $key ) ] = $value;
+				if ( isset( $value['expiration'] ) && $this->is_expired( $value['expiration'] ) ) {
+					$this->destroy( $key );
+				} else {
+					$tokens[ $key ] = $value;
+				}
 			}
 		}
 		return $tokens;
--- a/includes/class-token-user.php
+++ b/includes/class-token-user.php
@@ -203,4 +203,22 @@ class Token_User extends Token_Generic {
 		}
 		return update_user_meta( $this->user_id, $key, $info );
 	}
+
+	/**
+	 *
+	 */
+	public function find_token_users() {
+		$args     = array(
+			'count_total' => false,
+			'fields'      => 'ID',
+			'meta_query'  => array(
+				array(
+					'key'         => $this->prefix,
+					'compare_key' => 'LIKE',
+				),
+			),
+		);
+		$user_ids = get_users( $args );
+		return $user_ids;
+	}
 }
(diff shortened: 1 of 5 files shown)


dshanske:
@pfefferle What do you think about having this recurring job as a safety?

includes/class-token-user.php
```diff
@@ -145,32 +164,32 @@ public function get( $key, $hash = true ) {
...
+		$user_id = $results[0];
+		$value   = get_user_meta( $user_id, $key, true );
 		if ( empty( $value ) ) {
 			return false;
 		}
 
 		// If this token has expired destroy the token and return false;
 		if ( isset( $value['expiration'] ) && $this->is_expired( $value['expiration'] ) ) {
-			$this->destroy( $key, $user->ID );
+			$this->destroy( $key, $user_ID );
```
pfefferle: The `destroy()` function has no user_id param.

dshanske: Fixed that now.

dshanske:
@pfefferle Did everything else look good? I'm trying to work on reliability and stability as I incorporate new security measures.

indieauth.php
```diff
@@ -18,9 +18,55 @@
 	define( 'INDIEAUTH_REMOTE_MODE', 0 );
 }
 
+register_activation_hook( __FILE__, array( 'IndieAuth_Plugin', 'schedule' ) );
```
pfefferle: I would call it `activate` to match the `deactivate` case

dshanske: It's called on upgrade and when you load the config page as well, to check if the hook is scheduled. Usually, if it was only called on activation, I would call it activate.

pfefferle: but then we should also call the `deactivation` hook `deactivation`.

dshanske: Okay.

dshanske: Okay. Done.

indieauth.php
```diff
@@ -18,9 +18,55 @@
...
+	}
+
+	public static function deactivation() {
+		$timestamp = wp_next_scheduled( 'indieauth_cleanup', array( false ) );
+		if ( $timestamp ) {
+			wp_unschedule_event( $timestamp, 'indieauth_cleanup', array( false ) );
+		}
+	}
+
+	public static function schedule() {
```
pfefferle, line 60: rename to `activation`?

dshanske: But it isn't just called on activation. I can create a function that calls it called activation

dshanske: Done now.

pfefferle: ah, you added a new  function instead of renaming this one! nice 👍

pfefferle (approved):
LGTM!

(I only had a look at the code, but have not really tested the functionality)

dshanske:
I'm not going to push a new version till I add several additional improvements and checks. So I'll be doing more testing.

PR #195, opened 2021-03-01
fix C&P issue

pfefferle:
both tests use the same identifier, so you always get the header informations, even if you click on the `https` test results


dshanske (approved):
Oops

pfefferle:
Thanks :)

PR #198, opened 2021-03-26
Clean Up Strings

dshanske:
i18n guidelines state you shouldn't have HTML markup inside the string. I've rewritten this to place the HTML outside of the string, among other changes related to that style.

### DIFF
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -145,7 +145,19 @@ class IndieAuth_Authorization_Endpoint {
 	public static function scope_list( $scopes ) {
 		if ( ! empty( $scopes ) ) {
 			foreach ( $scopes as $s ) {
-				printf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), esc_html( self::scopes( $s ) ) ); // phpcs:ignore
+				echo wp_kses(
+					sprintf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), esc_html( self::scopes( $s ) ) ),
+					array(
+						'li'     => array(),
+						'strong' => array(),
+						'input'  => array(
+							'type'    => array(),
+							'name'    => array(),
+							'value'   => array(),
+							'checked' => array(),
+						),
+					)
+				);
 			}
 		}
 	}
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -145,7 +145,7 @@ class IndieAuth_Authorization_Endpoint {
 	public static function scope_list( $scopes ) {
 		if ( ! empty( $scopes ) ) {
 			foreach ( $scopes as $s ) {
-				printf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), self::scopes( $s ) );
+				printf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), esc_html( self::scopes( $s ) ) ); // phpcs:ignore
 			}
 		}
 	}
@@ -279,7 +279,7 @@ class IndieAuth_Authorization_Endpoint {
 	public function authorize() {
 		$current_user = wp_get_current_user();
 		// phpcs:disable
-		$client_id     = wp_unslash( $_GET['client_id'] ); // WPCS: CSRF OK
+		$client_id     = esc_url_raw( wp_unslash( $_GET['client_id'] ) ); // WPCS: CSRF OK
 		$info = new IndieAuth_Client_Discovery( $client_id );
 		$client_name = $info->get_name();
 		$client_icon = $info->get_icon();
@@ -290,11 +290,11 @@ class IndieAuth_Authorization_Endpoint {
 		}
 
 		$redirect_uri  = isset( $_GET['redirect_to'] ) ? wp_unslash( $_GET['redirect_to'] ) : null;
-		$scope         = isset( $_GET['scope'] ) ? wp_unslash( $_GET['scope'] ) : null;
+		$scope         = isset( $_GET['scope'] ) ? sanitize_text_field( wp_unslash( $_GET['scope'] ) ) : null;
 		$scopes        = array_filter( explode( ' ', $scope ) );
 		$state         = isset( $_GET['state'] ) ? $_GET['state'] : null;
-		$me            = isset( $_GET['me'] ) ? wp_unslash( $_GET['me'] ) : null;
-		$response_type = isset( $_GET['response_type'] ) ? wp_unslash( $_GET['response_type'] ) : null;
+		$me            = isset( $_GET['me'] ) ? esc_url_raw( wp_unslash( $_GET['me'] ) ) : null;
+		$response_type = isset( $_GET['response_type'] ) ? sanitize_text_field( wp_unslash( $_GET['response_type'] ) ) : null;
 		$code_challenge = isset( $_GET['code_challenge'] ) ? wp_unslash( $_GET['code_challenge'] ) : null;
 		$code_challenge_method = isset( $_GET['code_challenge_method'] ) ? wp_unslash( $_GET['code_challenge_method'] ) : null;
 
--- a/includes/class-indieauth-authorize.php
+++ b/includes/class-indieauth-authorize.php
@@ -118,8 +118,8 @@ abstract class IndieAuth_Authorize {
 			return;
 		}
 		if ( is_author() || is_front_page() ) {
-			wp_kses( sprintf( '<link rel="authorization_endpoint" href="%s" />' . PHP_EOL, $auth ), $kses );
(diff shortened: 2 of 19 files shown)


templates/indieauth-notices.php
```diff
@@ -0,0 +1,16 @@
+<div class="notices">
+	<?php
+	if ( wp_parse_url( $client_id, PHP_URL_HOST ) !== wp_parse_url( $redirect_uri, PHP_URL_HOST ) ) {
+		?>
+	<p class="redirect">
+		<strong> ⚠️  <?php esc_html_e( 'The redirect URL this app is using does not match the domain of the client ID.', 'indieauth' ); ?> </strong>
```
pfefferle, line 6: two spaces after the icon?

templates/indieauth-notices.php
```diff
@@ -0,0 +1,16 @@
...
+	</p>
+		<?php
+	}
+	if ( ! is_null( $code_challenge ) && 'S256' === $code_challenge_method ) {
+		?>
+	<p class="pkce">
+		<strong>🔒<?php esc_html_e( 'This app is using PKCE for security.', 'indieauth' ); ?> </strong>
```
pfefferle: no spaces after the icon?

templates/indieauth-authorize-form.php
```diff
@@ -1,52 +1,41 @@
...
+			<img src="<?php echo esc_url( $client_icon ); ?>" />
 		<?php } ?>
+		<strong> 
 		<?php
-			printf(
-				__( '<strong>%1$s wants to access your site.</strong>', 'indieauth' ), 
-				$client
-			);
+			/* translators: 1. Client */
+			printf( esc_html__( '%1$s wants to access your site.', 'indieauth' ), $client ); // phpcs:ignore
```
akirk: This doesn't need to be phpcs:ignored. It can be escaped like this without warning:
```suggestion
			echo esc_html( sprintf( __( '%1$s wants to access your site.', 'indieauth' ), $client ) );
```

includes/class-indieauth-authorization-endpoint.php
```diff
@@ -145,7 +145,7 @@ public static function scopes( $scope = 'all' ) {
 	public static function scope_list( $scopes ) {
 		if ( ! empty( $scopes ) ) {
 			foreach ( $scopes as $s ) {
-				printf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), self::scopes( $s ) );
+				printf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), esc_html( self::scopes( $s ) ) ); // phpcs:ignore
```
akirk: This doesn't need to be phpcs:ignored. It can be escaped like this without warning:
```suggestion
				echo wp_kses(
					sprintf( '<li><input type="checkbox" name="scope[]" value="%1$s" %2$s /><strong>%1$s</strong> - %3$s</li>', $s, checked( true, true, false ), self::scopes( $s ) ),
					array(
						'li'     => array(),
						'strong' => array(),
						'input'  => array(
							'type'    => array(),
							'name'    => array(),
							'value'   => array(),
							'checked' => array(),
						),
					)
				);
```

templates/indieauth-authenticate-form.php
```diff
@@ -1,49 +1,37 @@
...
-				$current_user->display_name,
-				$current_user->user_nicename
-				
+				/* translators: 1. Client with link 2. User ID 3. User Display Name 4. User Nicename */
+				'<p>' . esc_html__( 'The app %1$s would like to identify you as %2$s, which is user %3$s(%4$s).', 'indieauth' ) . '</p>',
+				$client, // phpcs:ignore
+				'<strong>' . esc_url( $user_website ) . '</strong>',
+				'<strong>' . esc_html( $current_user->display_name ) . '</strong>',
+				esc_html( $current_user->user_nicename )
 			);
```
akirk: This doesn't need to be phpcs:ignored. A possible escaped solution (mind that because of Github limitations this doesn't result in correct code, the prepending `<?php printf` needs to be replaced as well) would look like this:
```suggestion
		<p>
		<?php
			echo wp_kses(
				sprintf(
				/* translators: 1. Client with link 2. User ID 3. User Display Name 4. User Nicename */
					__( 'The app %1$s would like to identify you as %2$s, which is user %3$s (%4$s).', 'indieauth' ),
					$client,
					'<strong>' . esc_url( $user_website ) . '</strong>',
					'<strong>' . $current_user->display_name . '</strong>',
					$current_user->user_nicename
				),
				array( 'strong' => array() )
			);
			?>
		</p>
```

akirk (commented):
Hi! I've made a few suggestions around places where a `phpcs:ignore` was added to a changed line. In these cases I found that the escaping didn't go quite far enough and I've suggested to code to implement it in a way that ensures escaping of the variables and makes phpcs happy at the same time.

I also found a few more `phpcs:ignore` in the code base that could be removed in the scope of this PR:
- https://github.com/indieweb/wordpress-indieauth/blob/trunk/includes/class-indieauth-admin.php#L164
- https://github.com/indieweb/wordpress-indieauth/blob/trunk/includes/class-indieauth-authorize.php#L115
- https://github.com/indieweb/wordpress-indieauth/blob/trunk/includes/class-indieauth-token-ui.php#L66

dshanske:
@akirk Thanks, will update. Clearly there are opportunities here to improve the i18n markup without tripping the linting checks.

dshanske:
I went through all of them. Hopefully that is all.

dshanske:
@pfefferle Can you have a look and see if I satisfied?

includes/class-indieauth-authorize.php
```diff
@@ -108,12 +108,18 @@ public function http_header() {
...
+				'rel'  => array(),
+			),
+		);
 		if ( empty( $auth ) || empty( $token ) ) {
 			return;
 		}
 		if ( is_author() || is_front_page() ) {
-			printf( '<link rel="authorization_endpoint" href="%s" />' . PHP_EOL, $auth ); // phpcs:ignore
-			printf( '<link rel="token_endpoint" href="%s" />' . PHP_EOL, $token ); //phpcs:ignore
+			wp_kses( sprintf( '<link rel="authorization_endpoint" href="%s" />' . PHP_EOL, $auth ), $kses );
```
akirk: This won't output anything, so should be this:
```suggestion
			echo wp_kses( sprintf( '<link rel="authorization_endpoint" href="%s" />' . PHP_EOL, $auth ), $kses );
```

includes/class-indieauth-authorize.php
```diff
@@ -108,12 +108,18 @@ public function http_header() {
...
+			wp_kses( sprintf( '<link rel="token_endpoint" href="%s" />' . PHP_EOL, $token ), $kses );
```
akirk: ditto
```suggestion
			echo wp_kses( sprintf( '<link rel="token_endpoint" href="%s" />' . PHP_EOL, $token ), $kses );
```

templates/indieauth-settings.php
```diff
@@ -3,70 +3,78 @@
...
+						<?php esc_html_e( 'Use IndieAuth login', 'indieauth' ); ?>
 					</th>
 					<td>
 						<label for="indieauth_show_login_form">
-							<input type="checkbox" name="indieauth_show_login_form" id="indieauth_show_login_form" value="1" <?php
-								echo checked( true, get_option( 'indieauth_show_login_form' ) );  ?> />
-							<?php _e( 'Add a link to the login form to authenticate using an IndieAuth endpoint.', 'indieauth' ); ?>
+							<input type="checkbox" name="indieauth_show_login_form" id="indieauth_show_login_form" value="1" 
+							<?php
+								echo checked( true, get_option( 'indieauth_show_login_form' ) );
```
akirk, line 74: Nit but this could also just be:
```suggestion
								echo checked( get_option( 'indieauth_show_login_form' ) );
```

templates/websignin-form.php
```diff
@@ -1,24 +1,24 @@
...
+$login_errors = new WP_Error();
 login_header(
 	__( 'Sign in with your website', 'indieauth' ),
 	'',
-	$errors
+	$login_errors
 );
 ?>
-<form name="loginform" id="loginform" action="<?php add_query_arg( 'action', 'websignin', wp_login_url() ); ?>" method="post">
+<form name="loginform" id="loginform" action="<?php esc_url( add_query_arg( 'action', 'websignin', wp_login_url() ) ); ?>" method="post">
```
akirk: This would output nothing, so an echo is missing:
```suggestion
<form name="loginform" id="loginform" action="<?php echo esc_url( add_query_arg( 'action', 'websignin', wp_login_url() ) ); ?>" method="post">
```

akirk (commented):
I took another look and only found a few small remaining things.

dshanske:
@akirk Thanks for this...I tried to get all of them in one go. Will update and ask if you can have a look again. Cleaning up strings isn't the most exciting PR, but it needs to be done

dshanske:
@akirk @pfefferle I want to finally get this PR merged and the new version pushed, if you could have a look

templates/indieauth-settings.php
akirk: Actually this should be just 
```suggestion
								checked( get_option( 'indieauth_show_login_form' ) );
```
since the function is defined as `checked( mixed $checked, mixed $current = true, bool $echo = true )`

akirk (approved):
Looks good to me now.

pfefferle (approved):
LGTM

thanks @akirk for your review!

PR #205, opened 2021-05-20
Fix notice: Don't run JSON through wp_kses

akirk:
After #198 was applied, I get this message in wp-admin:
```
 Notice: Trying to get property 'message' of non-object in wp-content/plugins/indieauth/includes/class-indieauth-admin.php on line 220
```

This is because JSON is run through wp_kses. This brings back direct output of the JSON.


pfefferle (approved):
Makes sense!

PR #206, opened 2021-05-21
Use single quotes for auth test html attrs

chee:
should fix #204, close #201 and fix #188 and close #158

PR #208, opened 2021-05-25
Sanitize Fix

dshanske:
This fixes the issue identified in #206 by escaping the HTML, but then using htmlspecialchars to restore it for display purposes.

It also bumps the version so this can be released.


dshanske:
It also changes the test token to a token the length of a standard token, generated using the same code, so if any mod_security might be tripped, this can be tested.

pfefferle:
But the problem with the "broken" JSON message (mentioned by @chee ) is still existent: https://github.com/indieweb/wordpress-indieauth/pull/206#issuecomment-846208896

pfefferle (approved):
LGTM

PR #210, opened 2021-06-13
Add expiration functionality

dshanske:
This adds expiration functionality. By default, all tokens will expire within two weeks. It can be changed in the settings screen.

The bulk action functionality in the token admin can renew or disable expiry. The manual add token can set a custom expiring token. The existing cron job will clean up expired tokens.

The option to expire all tokens older than Y is now removed as unnecessary.

### DIFF
--- a/includes/class-indieauth-token-ui.php
+++ b/includes/class-indieauth-token-ui.php
@@ -59,8 +59,9 @@ class IndieAuth_Token_UI {
 		if ( empty( $scopes ) ) {
 			$scopes = 'create update';
 		}
-		$scopes = sanitize_text_field( $scopes );
-		$token  = self::generate_local_token( $client_name, $scopes );
+		$scopes  = sanitize_text_field( $scopes );
+		$expires = sanitize_text_field( $_REQUEST['expires_in'] );
+		$token   = self::generate_local_token( $client_name, $scopes, $expires );
 		?>
 	<p><?php esc_html_e( 'A token has been generated and appears below. This token will not be stored anywhere. Please copy and store it.', 'indieauth' ); ?></p>
 	<h3><?php echo $token; // phpcs:ignore 
@@ -71,7 +72,7 @@ class IndieAuth_Token_UI {
 		exit;
 	}
 
-	private function generate_local_token( $name, $scopes ) {
+	private function generate_local_token( $name, $scopes, $expires_in = 0 ) {
 		$user_id = get_current_user_id();
 		$tokens  = new Token_User( '_indieauth_token_' );
 		$tokens->set_user( $user_id );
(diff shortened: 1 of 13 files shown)


includes/class-indieauth-token-ui.php
```diff
@@ -59,8 +59,9 @@ public function new_token() {
 		if ( empty( $scopes ) ) {
 			$scopes = 'create update';
 		}
-		$scopes = sanitize_text_field( $scopes );
-		$token  = self::generate_local_token( $client_name, $scopes );
+		$scopes  = sanitize_text_field( $scopes );
+		$expires = sanitize_text_field( $_REQUEST['expires_in'] );
```
pfefferle, line 63: `$_REQUEST['expires_in']` should be sanitized.

pfefferle, line 63: Can `$_REQUEST['expires_in']` be empty or unset?

readme.md
```diff
@@ -4,7 +4,7 @@
 **Requires at least:** 4.9.9  
 **Requires PHP:** 5.6  
 **Tested up to:** 5.7  
-**Stable tag:** 3.6.2  
+**Stable tag:** 3.7.0
```
pfefferle: If `expires = on` is the default behaviour, I would recommend to use 4.0.0 as version, because it is a breaking change.

dshanske: Agreed. Also will add the option to restore the original behavior just in case

includes/class-indieauth-token-ui.php
dshanske: I originally was going to add the option to the authorize form, but didn't....so this line is unnecessary...but I may add that feature.

pfefferle:
There is a PHPCS error that also blocks the merge.

dshanske:
Updated. This adds in the option to disable the expiry overall by setting 0. It also changes the new release version to 4.0.0.

includes/class-token-generic.php
```diff
@@ -31,13 +31,17 @@ public function renew( $key ) {
...
+		// Ignore if the renewal time is 0.
+		if ( 0 === $expires ) {
+			return;
+		}
 		if ( array_key_exists( 'expiration', $token ) ) {
-			$token['expiration'] = $token['expiration'] + get_option( 'indieauth_expires_in' );
+			$token['expiration'] = $token['expiration'] + $expires;
 		} else {
-			$token['expiration'] = time() + get_option( 'indieauth_expires_in' );
+			$token['expiration'] = time() + $xpires;
```
pfefferle: there is an `e` missing. should be `$expires` instead of `$xpires`

dshanske: Fixed.

includes/class-token-list-table.php
```diff
@@ -20,12 +20,9 @@ public function get_columns() {
 
 	public function get_bulk_actions() {
 			  return array(
```
pfefferle: indents!

templates/indieauth-authorize-form.php
```diff
@@ -47,6 +47,18 @@
 		<?php self::scope_list( $scopes ); ?>
 		</ul>
 	</div>
+	<div class="expiration">
+		<?php
+			$expiration = (int) get_option( 'indieauth_expires_in' );
```
pfefferle: indents!

pfefferle (approved):

PR #212, opened 2021-07-10
Ticket Endpoint

dshanske:
This merge implements an experimental ticket endpoint to receive tickets. It adds a UI to revoke or verify those tickets. The endpoint is disabled by default and only enabled with a wp-config flag.

### DIFF
--- a/includes/class-external-token.php
+++ b/includes/class-external-token.php
@@ -1,0 +1,223 @@
+<?php
+
+/* Class for managing external tokens in user meta */
+class External_User_Token {
+
+	private $user_id;
+
+	public function __construct( $user_id = null ) {
+		if ( is_numeric( $user_id ) ) {
+			$this->user_id = $user_id;
+		} else {
+			$this->user_id = get_current_user_id();
+		}
+	}
+
+	/**
+	 * Retrieves a token
+	 *
+	 * @param string  $key token to retrieve.
+	 * @return array|boolean Token data or false if not found
+	 */
+	public function get( $key ) {
+		if ( ! is_string( $key ) ) {
+			return false;
+		}
+
+		if ( ! current_user_can( 'edit_user', $this->user_id ) ) {
+			return false;
+		}
+
+		$tokens = get_user_meta( $this->user_id, 'indieauth_external_tokens', true );
+		foreach ( $tokens as $key => $token ) {
+			if ( in_array( $token['access_token'], $key, true ) ) {
+				return $tokens[ $key ];
+			}
+		}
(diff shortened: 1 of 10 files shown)


dshanske:
@pfefferle Can you have a look at this?

includes/class-indieauth-debug.php
```diff
@@ -31,7 +31,7 @@ public function indieauth_allow_localhost( $r, $url ) {
 	public function log_rest_api_errors( $result, $server, $request ) {
 		$request_route = $request->get_route();
 		$result_route  = $result->get_matched_route();
-		$routes        = array( 'indieauth/1.0/auth', '/indieauth/1.0/token' );
+		$routes        = array( 'indieauth/1.0/auth', '/indieauth/1.0/token', 'indieauth/1.0/ticket' );
```
tw2113, line 34: Won't be a showstopper, but this feels like it may be getting primed for a dedicated method for retrieving routes. Could also potentially get a filter for 3rd party integration.

dshanske: Good point. This was written a long time ago. I should have it retrieve all indieauth namespace routes.

includes/class-external-token.php
```diff
@@ -0,0 +1,314 @@
...
+		return $resp;
+	}
+
+
+	/**
+	 * Revokes an External token.
+	 *
+	 * @param array $key Token to verify.
+	 * @param boolean $revoke Whether to send revoke request to token endpoint.
+	 * @return boolean|array Either false or the response from the token endpoint.
```
tw2113, line 261: Minor nitpick, but this feels like it doesn't revoke, just verify. Plus there's the `$revoke` parameter that doesn't exist on this `verify()` method

includes/class-indieauth-debug.php
dshanske: I'll likely save the idea for a future PR though.

includes/class-external-token.php
dshanske: Fixing that now. I think I just used the docblock from another function and forgot to fix it.

tw2113:
Nothing notable standing out to me with this one, code wise. Feels all like pretty well known HTTP process and workflow which is good.

dshanske:
> Nothing notable standing out to me with this one, code wise. Feels all like pretty well known HTTP process and workflow which is good.

Great, appreciate the help. It's still an experimental spec, but I'd like to merge it in so I can iterate on it.

tw2113 (approved):

PR #213, opened 2021-07-11
Turn create into a create/draft/none option selection

dshanske:
This changes the authorize screen to allow you to dynamically change a create scope request into a draft scope. This means even clients that don't support draft can be given a token that will only create draft posts. This is a visual only change.


dshanske:
@pfefferle Can you have a look?

tw2113 (approved):

PR #217, opened 2021-09-08
Introspection

dshanske:
As per https://github.com/indieweb/indieauth/issues/33, we're adopting the response used by the Token Introspection specification. This means the properties I used, which were internal only, 'expiration' and 'issued_at' should be changed to 'iat' and 'exp' to match. While I could just change them on the response, there was no reason to retain the old structure. 

I also changed expires_in to no longer be stored. It is generated only on return, as there is no need to store it.

While the return parameters of the GET token verification are amended, it still, like the original spec, returns a 400 series error for backward compatibility. The new introspection endpoint, which is a POST request, always returns a 200. At some point when clients have caught up, will likely adjust accordingly. 

There is a slight reformatting of the POST action handler, which now uses two switch statements, one for action, the other for grant_type=. The reason for this is somewhat practical. 

For grant_type, we only support 1 right now, but the new amendment decided at the popup was to include refresh_tokens as a recommendation in the spec(https://github.com/indieweb/indieauth/issues/81) so there would be a new grant_type, which I intend to submit as a separate PR.  In https://github.com/indieweb/indieauth/issues/87, @Zegnat proposes extending action as part of the Ticket Auth proposal, so might as well structure for that as well.

### DIFF
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -241,11 +241,11 @@ class IndieAuth_Authorization_Endpoint {
 			return new WP_OAuth_Response( 'invalid_grant', __( 'Invalid authorization code', 'indieauth' ), 400 );
 		}
 		$user = get_user_by( 'id', $token['user'] );
-		if ( $token['expiration'] <= time() ) {
+		if ( $token['exp'] <= time() ) {
 			$this->delete_code( $code, $token['user'] );
 			return new WP_OAuth_Response( 'invalid_grant', __( 'The authorization code expired', 'indieauth' ), 400 );
 		}
-		unset( $token['expiration'] );
+		unset( $token['exp'] );
 		// If there is a code challenge
 		if ( isset( $token['code_challenge'] ) ) {
 			$code_verifier = $request->get_param( 'code_verifier' );
--- a/includes/class-indieauth-local-authorize.php
+++ b/includes/class-indieauth-local-authorize.php
@@ -35,11 +35,11 @@ class IndieAuth_Local_Authorize extends IndieAuth_Authorize {
 			return $return;
 		}
 		$return['last_accessed'] = time();
-		if ( array_key_exists( 'expiration', $return ) ) {
-			$return['expires_in'] = $return['expiration'] - time();
-		}
 		$return['last_ip'] = $_SERVER['REMOTE_ADDR'];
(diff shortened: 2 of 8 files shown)


tw2113 (approved):
All looks good to me.

includes/class-indieauth-authorization-endpoint.php
```diff
@@ -241,11 +241,11 @@ public function verify( $request ) {
 			return new WP_OAuth_Response( 'invalid_grant', __( 'Invalid authorization code', 'indieauth' ), 400 );
 		}
 		$user = get_user_by( 'id', $token['user'] );
-		if ( $token['expiration'] <= time() ) {
+		if ( $token['exp'] <= time() ) {
```
jamietanna, line 244: :question: is this the response from the token endpoint?

includes/class-indieauth-local-authorize.php
```diff
@@ -35,11 +35,11 @@ public function verify_access_token( $token ) {
 			return $return;
 		}
 		$return['last_accessed'] = time();
-		if ( array_key_exists( 'expiration', $return ) ) {
-			$return['expires_in'] = $return['expiration'] - time();
-		}
-		$return['last_ip'] = $_SERVER['REMOTE_ADDR'];
+		$return['last_ip']       = $_SERVER['REMOTE_ADDR'];
 		$tokens->update( $token, $return );
+		if ( array_key_exists( 'exp', $return ) ) {
```
jamietanna, line 40: :wrench: to reduce the diff noise this could be moved up to where it was

jamietanna (commented):
Implementation looks good - main thing I'm wary of is the move of the `expiration` to `exp` in the token, which looks like a breaking change that'd invalidate existing tokens?

includes/class-indieauth-authorization-endpoint.php
dshanske: No. This is the authorization endpoint. I use the same expiration code for authorization codes as I do for tokens, so I standardized using exp.

dshanske: I'm not returning it.

dshanske:
> Implementation looks good - main thing I'm wary of is the move of the `expiration` to `exp` in the token, which looks like a breaking change that'd invalidate existing tokens?

It checks for the old parameter and uses it as well, for the migration.

PR #221, opened 2021-09-13
Add Additional Tests and Fixes Resulting From Testing

dshanske:
This adds a variety of new tests, and makes some fixes based on the results.  Had originally proposed recent changes be 4.0.1, however, with the number of changes, this should be a 4.1.0, so updated version pending release.

### DIFF
--- a/includes/class-indieauth-client-discovery.php
+++ b/includes/class-indieauth-client-discovery.php
@@ -9,7 +9,12 @@ class IndieAuth_Client_Discovery {
 
 	public function __construct( $client_id ) {
 		$this->client_id = $client_id;
-		$this->html      = self::parse( $client_id );
+
+		if ( defined( 'INDIEAUTH_UNIT_TESTS' ) ) {
+			return;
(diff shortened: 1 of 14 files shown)


includes/class-indieauth-client-discovery.php
```diff
@@ -9,7 +9,12 @@ class IndieAuth_Client_Discovery {
 
 	public function __construct( $client_id ) {
 		$this->client_id = $client_id;
-		$this->html      = self::parse( $client_id );
+
+		if ( defined( 'INDIEAUTH_UNIT_TESTS' ) ) {
+			return;
+		}
+
+		$this->html = self::parse( $client_id );
```
tw2113, line 17: Feels slightly odd to set the client_id property and then pass in the already passed in $client_id from the constructor, instead of perhaps relying on the newly set property.

tw2113 (approved):
Only had that one spot that felt a bit off, but not a blocker.

PR #223, opened 2021-09-19
Refresh Tokens

dshanske:
Introduce Refresh Tokens and what came with it.

Introduce uuid into the auth code, so that if by some chance the auth code isn't revoked, we can tie it to the right access token. UUID for refresh token is the same as the access token so that we can revoke appropriately. Introduction of find by uuid function with testing.

Also, found two minor bugs introduced in previous PRs approved that are now fixed.

### DIFF
--- a/includes/class-indieauth-token-endpoint.php
+++ b/includes/class-indieauth-token-endpoint.php
@@ -7,8 +7,11 @@
 
 class IndieAuth_Token_Endpoint {
 	private $tokens;
+	private $refresh_tokens;
 	public function __construct() {
-		$this->tokens = new Token_User( '_indieauth_token_' );
+		$this->tokens         = new Token_User( '_indieauth_token_' );
+		$this->refresh_tokens = new Token_User( '_indieauth_refresh_' );
+
 		add_action( 'rest_api_init', array( $this, 'register_routes' ) );
 	}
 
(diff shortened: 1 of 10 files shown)


dshanske:
Including @jamietanna as a previous refresh token implementer to compare implementations, and @tw2113 as a WordPress check.

tw2113 (approved):

includes/class-indieauth-token-endpoint.php
```diff
@@ -119,16 +122,47 @@ public function get( $request ) {
...
+	 * Sets a refresh token based on an access token.
+	 * @param array $token Access Token Return.
+	 * @param int $user User ID.
+	 * @return string Refresh Token.
+	 */
+	public function set_refresh_token( $token, $user ) {
+		$refresh = array(
+			'scope'     => $token['scope'],
+			'client_id' => $token['client_id'],
+			'iat'       => time(),
```
jamietanna, line 153: looks like we don't use `exp` as the expiry is stored separately?

includes/class-indieauth-token-endpoint.php
```diff
@@ -192,56 +228,89 @@ public function post( $request ) {
...
+		if ( ! empty( $diff ) ) {
+			return new WP_OAuth_Response( 'invalid_request', __( 'The request is missing one or more required parameters', 'indieauth' ), 400 );
+		}
+		$refresh = $this->refresh_tokens->get( $params['refresh_token'] );
+		if ( ! $refresh ) {
+			return new WP_OAuth_Response( 'invalid_grant', __( 'Invalid Token', 'indieauth' ), 400 );
+		}
+
+		// Destroy the refresh token.
+		$this->refresh_tokens->destroy( $params['refresh_token'] );
```
jamietanna, line 245: This may be happening too early, i.e. if the below fails?

includes/class-indieauth-token-endpoint.php
```diff
@@ -192,56 +228,89 @@ public function post( $request ) {
...
+
+		if ( array_key_exists( 'client_id', $params ) && ( $refresh['client_id'] !== $params['client_id'] ) ) {
```
jamietanna: I think this isn't required - it's something I've mis-implemented on my side (confusing the request with how the credentialed clients work)

includes/class-indieauth-token-endpoint.php
```diff
@@ -192,56 +228,89 @@ public function post( $request ) {
...
 			array(
 				'code'          => $params['code'],
 				'redirect_uri'  => $params['redirect_uri'],
 				'client_id'     => $params['client_id'],
 				'code_verifier' => isset( $params['code_verifier'] ) ? $params['code_verifier'] : null,
 			)
 		);
-		$response = indieauth_verify_local_authorization_code( $args );
+		$response              = indieauth_verify_local_authorization_code( $args );
+		$response['client_id'] = $params['client_id'];
```
jamietanna: is this not stored as part of the authorization code?

includes/class-token-user.php
```diff
@@ -245,4 +245,21 @@ public function find_token_users( $refresh = false ) {
...
 		return $user_ids;
 	}
+
+	/**
+	 * Find tokens by UUID.
+	 *
+	 * @param string $uuid UUID.
+	 * @return array Results.
+	 */
+	public function find_by_uuid( $uuid, $user = null ) {
```
jamietanna: Could we use a more meaningful name? (Here and elsewhere)

```suggestion
	public function find_by_consent_id( $uuid, $user = null ) {
```

languages/indieauth.pot
```diff
@@ -5,7 +5,7 @@ msgstr ""
 "Project-Id-Version: IndieAuth 4.1.0\n"
 "Report-Msgid-Bugs-To: "
 "https://wordpress.org/support/plugin/wordpress-indieauth\n"
-"POT-Creation-Date: 2021-09-13 02:58:46+00:00\n"
+"POT-Creation-Date: 2021-09-19 04:08:37+00:00\n"
```
jamietanna, line 8: Did this need to change?

includes/class-indieauth-token-endpoint.php
dshanske: It's added when the token is saved.

includes/class-indieauth-token-endpoint.php
dshanske: If it fails, shouldn't it require the full flow?

tests/test-token-endpoint.php
```diff
@@ -191,6 +223,36 @@ public function test_auth_code_redemption_with_just_profile() {
 		static::$test_auth_code['scope'] = 'create';
 	}
 
+	// Sets a Refresh Token Then Redeems it Using the Grant Type.
```
jamietanna, line 226:
```suggestion
	    // Sets a Refresh Token Then Redeems it Using the Grant Type.
```

includes/class-indieauth-token-endpoint.php
dshanske: Well, the client id is optional, so I debated not checking.

includes/class-indieauth-token-endpoint.php
dshanske: It is.

languages/indieauth.pot
dshanske: It's part of the automated build process. Didn't want to do another PR just to rebuild it.

dshanske: It's part of the automated build process. Didn't want to do another PR just to rebuild it.

PR #226, opened 2021-11-26
Add IndieAuth Metadata Endpoint

dshanske:
This adds a metadata endpoint, as merged into the spec, into the plugin. It also adds the issuer URL to the authorization response return, also merged into the spec.

As a side effect, some other minor tweaks were necessary. The rest_index functionality is very similar to metadata, so it moved over to that file.

Need someone to check the code on a WordPress level, and on a new spec level.

### DIFF
--- a/includes/class-indieauth-metadata-endpoint.php
+++ b/includes/class-indieauth-metadata-endpoint.php
@@ -1,0 +1,118 @@
+<?php
+/**
+ * Metadata Endpoint Functionality
+ */
+class IndieAuth_Metadata_Endpoint {
+
+	public function __construct() {
+		add_filter( 'rest_index', array( $this, 'register_index' ) );
+		add_action( 'rest_api_init', array( $this, 'register_routes' ) );
+		add_action( 'wp_head', array( $this, 'html_header' ) );
+		add_action( 'template_redirect', array( $this, 'http_header' ) );
+	}
+
+	/**
+	 * Returns the URL for the metadata endpoint.
+	 **/
+	public static function get_metadata_endpoint() {
+		return rest_url( '/indieauth/1.0/metadata' );
+	}
+
+	/**
+	 * Add authentication information into the REST API Index
+	 *
+	 * @param WP_REST_Response $response REST API Response Object
+	 *
+	 * @return WP_REST_Response Response object with endpoint info added
+	 **/
+	public function register_index( WP_REST_Response $response ) {
+		$data      = $response->get_data();
+		$endpoints = array(
+			'authorization' => indieauth_get_authorization_endpoint(),
+			'token'         => indieauth_get_token_endpoint(),
+			'metadata'      => indieauth_get_metadata_endpoint(),
+		);
+		$endpoints = array_filter( $endpoints );
+		if ( empty( $endpoints ) ) {
+			return $response;
+		}
+		$data['authentication']['indieauth'] = array(
+			'endpoints' => $endpoints,
+		);
+		$response->set_data( $data );
+		return $response;
+	}
+
+	public function http_header() {
+		if ( is_author() || is_front_page() ) {
+			header( sprintf( 'Link: <%s>; rel="indieauth-metadata"', static::get_metadata_endpoint() ), false );
+		}
+	}
+	public static function html_header() {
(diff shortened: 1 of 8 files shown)


includes/class-indieauth-metadata-endpoint.php
```diff
@@ -0,0 +1,118 @@
...
+			'response_types_supported'         => array( 'code' ),
+			'grant_types_supported'            => $grants,
+			'service_documentation'            => 'https://indieauth.spec.indieweb.org',
+			'token_endpoint'                   => indieauth_get_token_endpoint(),
+			'code_challenge_methods_supported' => array( 'S256' ),
+			'authorization_response_iss_parameter_supported' => true,
+		);
+
+		$metadata = apply_filters( 'indieauth_metadata', $metadata );
+		return new WP_REST_Response(
```
Ruxton, line 109: should there be some kind of try/catch around all the above to return a decent response when there's an error and the response ISN'T 200?

includes/class-indieauth-metadata-endpoint.php
```diff
@@ -0,0 +1,118 @@
...
+		return rest_url( '/indieauth/1.0/metadata' );
+	}
+
+	/**
+	 * Add authentication information into the REST API Index
+	 *
+	 * @param WP_REST_Response $response REST API Response Object
+	 *
+	 * @return WP_REST_Response Response object with endpoint info added
+	 **/
```
pfefferle, line 27: only one `*`

includes/class-indieauth-metadata-endpoint.php
```diff
@@ -0,0 +1,118 @@
...
+	public function __construct() {
+		add_filter( 'rest_index', array( $this, 'register_index' ) );
+		add_action( 'rest_api_init', array( $this, 'register_routes' ) );
+		add_action( 'wp_head', array( $this, 'html_header' ) );
+		add_action( 'template_redirect', array( $this, 'http_header' ) );
+	}
+
+	/**
+	 * Returns the URL for the metadata endpoint.
+	 **/
```
pfefferle: only one `*`

includes/class-indieauth-metadata-endpoint.php
```diff
@@ -0,0 +1,118 @@
...
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'metadata' ),
+					'args'                => array(),
+					'permission_callback' => '__return_true',
+				),
+			)
+		);
+	}
+
+	/*
```
pfefferle: here are two `*` required

readme.txt
```diff
@@ -143,6 +143,11 @@ Since the extension is developing, there is currently not a specified way to tra
 
 == Upgrade Notice ==
 
+- 4.2.0 =
+
+Changes in the 4.2.0 branch implement future breaking changes to IndieAuth. Backward compatibility will be maintained for the foreseeable future, but clients are advised to update to the latest version of the standard
+to take advantage of the latest opportunities.
```
Ruxton: I'm being super fickle, but it doesn't need a new line. ;P

Ruxton (approved):
looks good

includes/class-indieauth-metadata-endpoint.php
dshanske: What error are you envisioning?

includes/class-indieauth-metadata-endpoint.php
dshanske: Will fix, how is it overall?

pfefferle:
@dshanske is this version already breaking backwards compatibility?

readme.txt
pfefferle: me too ;)

pfefferle (approved):
LGTM

...but I haven't tested the functionality!

dshanske:
> @dshanske is this version already breaking backwards compatibility?

None of the changes I'm putting in for 4.2.0, which were discussed at the IndieAuth2 popup and subsequently, are breaking backward compatibility. I just wanted to put in the announcement that eventually, the various backward compatibility measures would be removed. Right now, there isn't any support for the new features, so no point in removing old.

PR #227, opened 2021-12-02
New Endpoints

dshanske:
This, as a continuation of the previous PR, introduces a dedicated revocation, introspection, and userinfo endpoint. These all separate functions previously in the single endpoint, now that the metadata endpoint allows this without having to overload it. This is reflected in current PRs discussed at the last popup awaiting merge. The Metadata endpoint was already merged, which is why the introspection endpoint is being adjusted appropriately, as that was changed after this decision.

It keeps the built-in functionality for now, until there is more adoption.


tw2113 (approved):
Looks good as best I can tell.

PR #231, opened 2021-12-25
Clients

dshanske:
Introduce a Taxonomy to store client data and relevant functions

### DIFF
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -380,9 +380,14 @@ class IndieAuth_Authorization_Endpoint {
 		$current_user = wp_get_current_user();
 		// phpcs:disable
 		$client_id     = esc_url_raw( wp_unslash( $_GET['client_id'] ) ); // WPCS: CSRF OK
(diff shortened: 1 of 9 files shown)


tw2113 (approved):

pfefferle:
Is it really a good idea to use a taxonomy to save client data? If you do not want to use custom tables I even would prefer to use custom post types.

dshanske:
@pfefferle I'm attaching it to posts made via Micropub. It allows for proper attribution of posts

dshanske:
I debated where to store it for a bit.

tw2113:
post to post isn't as straightforward for post content association as compared to taxonomy term/term meta and even then just post meta. At least term meta has some indexing available to it by way of terms. In the end it feels like a micro-optimization of where to store data for a given content type, and how much data is already generally queried for by default vs what would be extra.

Less a question of "best practice" in my mind, and more of "best preferred" question.

dshanske:
Still thinking about this.

pfefferle (approved):
I saw this more and more often lately... maybe this is really the way to go...

dshanske:
I have to go remind myself what I did, but might be time for a release.

PR #234, opened 2022-07-09
Updates GitHub Actions and Composer

dshanske:
This takes care of the new composer requirements and updated the GitHub Actions scripts to accommodate the latest versions. Also now allows for testing on PHP8.0


composer.json
```diff
@@ -27,7 +33,6 @@
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
     "sebastian/phpcpd": "^3.0 || ^4.0 || ^6.0",
-    "phpunit/phpunit": "^5.0 || ^6.0 || ^7.5",
```
pfefferle, line 30: Why removing phpunit?

dshanske: It's not needed, it's a dependency of the yoast polyfill. The polyfill recommends it, but we couldn't really use it till WordPress was updated to support it

dshanske: I also deployed it identically to other repositories...your PR I merged to jsonfeed for a GitHub action prompted me to update them all

pfefferle (approved):

PR #236, opened 2022-08-29
Fix #235

### LINKED ISSUE #235: indieauth_cleanup expects class Token_User
```
[21-Aug-2022 08:30:02 UTC] PHP Fatal error:  Uncaught Error: Class "Token_User" not found in <...>/public_html/wp-content/plugins/indieauth/indieauth.php:81
Stack trace:
#0 <...>/public_html/wp-includes/class-wp-hook.php(307): IndieAuth_Plugin::expires()
#1 <...>/public_html/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters()
#2 <...>/public_html/wp-includes/plugin.php(524): WP_Hook->do_action()
#3 <...>/public_html/wp-cron.php(138): do_action_ref_array()
#4 {main}
  thrown in <...>/public_html/wp-content/plugins/indieauth/indieauth.php on line 81
```

I recently switched to a remote endpoint, which is why I think `Token_User` is no longer autoloaded when this cron event is run. Maybe add an extra check in there? Or load it regardless, so that old tokens can still be cleaned up?

janboddez:
Seems to be as easy as this, i.e., _always_ require `class-token-user.php`. Don't think it affects test cases, although I have not been able to run them.

### DIFF
--- a/indieauth.php
+++ b/indieauth.php
@@ -97,6 +97,7 @@ class IndieAuth_Plugin {
 				'functions.php', // Global Functions
 				'class-oauth-response.php', // OAuth REST Error Class
 				'class-token-generic.php', // Token Base Class
+				'class-token-user.php',
 				'class-indieauth-scope.php', // Scope Class
 				'class-indieauth-scopes.php', // Scopes Class
 				'class-indieauth-authorize.php', // IndieAuth Authorization Base Class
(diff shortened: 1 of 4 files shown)


janboddez:
O, didn't know newer commits would be added tho. (I'm a Git noob.)

So, somewhat off-topic: there's what looks like a typo (missing space before the opening parenthesis in the auth form:

```
'The app %1$s would like to identify you as %2$s, which is user %3$s(%4$s).'
```
And:
```
'The app will use credentials of %1$s(%2$s). You can revoke access at any time.'
```

There _might_  be a few of these in `class-indieauth-debug.php`, too, unless the missing space serves a purpose somehow, and since end users don't normally get to see debug messages I did not include them.

dshanske (approved):
I don't think there is an issue in loading the Token User class always.

PR #239, opened 2022-12-21
Resolve Issue 237

### LINKED ISSUE #237: Fix Missing HTML Header for Metadata
https://github.com/indieweb/wordpress-indieauth/blob/87ce000d79518e67215b80841a2fb883125353b9/includes/class-indieauth-metadata-endpoint.php#L60

Not sure why this line is here, but it is removing the HTML header

dshanske:
This resolves #237 and bumps the version to release for the last few minor fixes, and to show WordPress 6.1 support.


pfefferle (approved):
LGTM

PR #245, opened 2023-04-09
Allow other OAuth2 providers

akirk:
If there is another OAuth2 provider plugin installed, IndieAuth overrules it and doesn't give it a chance to authenticate the user. This is because the `determine_current_user` filter doesn't act as a filter but makes a definite decision by returning `0`. This changes that behavior and adds tests for it.

### DIFF
--- a/includes/class-indieauth-authorize.php
+++ b/includes/class-indieauth-authorize.php
@@ -155,15 +155,16 @@ abstract class IndieAuth_Authorize {
 
 		$params = $this->verify_access_token( $token );
 		if ( ! isset( $params ) ) {
-			return 0;
+			return $user_id;
 		}
 		if ( is_oauth_error( $params ) ) {
-			$this->error = $params;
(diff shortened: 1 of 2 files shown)


dshanske:
So what happens if both auths fail?

pfefferle:
Then User Id is null and it fails.

akirk:
Like @pfefferle said. I can add another test for that case.

dshanske:
Yes, but I'm talking about error returning to the client.

akirk:
I added a test for when no valid token for another authentication provider is provided. When accessing via REST, `rest_authentication_errors` will still kick in, otherwise, if the resource can be accessed without authentication, an unauthenticated response will be returned.

dshanske (approved):
No issue with these changes at this time. It never occurred to me someone would want an OAuth2 provider and an IndieAuth provider simultaneously.

Issue #244, opened 2023-02-12
New Client Storage doesn't accept duplicate terms well


dshanske:
[A term with the name provided already exists in this taxonomy.] wants to access your site.

PR #250, opened 2023-09-02
Switch from using name as the slug to client ID

dshanske:
In testing the new taxonomy, using the name of the application is too fragile. This ensures unique slugs in the event of two clients with the same name.


dshanske:
Addresses #244

pfefferle:
Why not use `sanitize_title`? It makes `https-notiz-blog-about` out of `https://notiz.blog/about/`.

dshanske:
Wanted to strip the www and the https

pfefferle (approved):

PR #254, opened 2023-11-11
Fix deprecation notice

janboddez:
`rawurldecode()` does not or no longer (I'm on PHP 8.1 right now) accept `null` as a first param. As such, a deprecation warning occasionally pops up in WordPress' debug log (if it happens to be enabled, of course).

By setting it to an empty string instead, this is notice no longer appears.

Also, later on `wp_login_url( $redirect_to )` is run; I checked the [documentation for `wp_login_url()`](https://developer.wordpress.org/reference/functions/wp_login_url/) and found that the default value for its first argument is an empty string, and that it actually expects a _string_. I.e., I think a string is okay. :-)

### DIFF
--- a/includes/class-indieauth-client-taxonomy.php
+++ b/includes/class-indieauth-client-taxonomy.php
@@ -118,7 +118,7 @@ final class IndieAuth_Client_Taxonomy {
 		// strip leading www, if any.
 		$host = preg_replace( '/^www\./', '', $host );
 		$path = wp_parse_url( $url, PHP_URL_PATH );
-		$path = str_replace( '/', '_', $title );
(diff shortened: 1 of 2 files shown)


dshanske (approved):
Looks good to me.

janboddez:
Fixed another one, I think the `$title` var (which doesn't exist) on line 121 of class-indieauth-client-taxonomy.php should really be `$path`? Although I do hope this doesn't accidentally break anything else.

dshanske:
It wouldn't.. that it me thinking about generating slugs for post titles...which this isn't.

Actually working on a bunch of cleanup for this plugin right now...so anything you see can get rolled up into the 4.4 release I'm planning

Issue #267, opened 2024-06-15
Update Discovery with Support for New Proposed Discovery Method


dshanske:
https://github.com/indieweb/indieauth/issues/133

https://drafts.aaronpk.com/draft-parecki-oauth-client-id-metadata-document/draft-parecki-oauth-client-id-metadata-document.html

janboddez:
Ha, this must be why I can't log in to the IndieWeb wiki. Right?

Looks like Indielogin.com has already switched to JSON client metadata, and the plugin isn't yet compatible. https://github.com/aaronpk/indielogin.com/blob/35c8aa0ae627517d9c9a9578b901740736ded428/app/Controller.php#L18

Edit: just seeing the PR now. I'll try it when I can. Aaron's also right, the error handling can be improved. E.g., there _seem_ to be some checks in the code (`if ( ! $input ) ...` and the like), but they may be ineffective. Might be able to help out there, I'll see what I can do.

dshanske:
Merged via #268

PR #268, opened 2024-06-17
Add preliminary support for json metadata client discovery

dshanske:
#267 per request, basic support for the client ID being a JSON file using the properties identified.

@aaronpk This should work based on the notes.

If this gets adopted into the spec, I'll probably cut the unsupported manifest side file code, as that proposal never took off, and simplify for JSON with a HTML MF2 fallback.

### DIFF
--- a/includes/class-indieauth-client-discovery.php
+++ b/includes/class-indieauth-client-discovery.php
@@ -2,7 +2,6 @@
 
 class IndieAuth_Client_Discovery {
 	protected $rels     = array();
-	protected $manifest = array();
 	protected $html     = array();
 	protected $mf2      = array();
 	protected $json     = array();
@@ -25,7 +24,11 @@ class IndieAuth_Client_Discovery {
 		);
 
 		// If this is an IP address on the donotfetch list then do not fetch.
-		if ( ( $ip && ! in_array( $ip, $donotfetch, true ) || 'localhost' === wp_parse_url( $client_id, PHP_URL_HOST ) ) ) {
+		if ( $ip && ! in_array( $ip, $donotfetch, true ) ) {
+			return;
+		}
+
+		if ( 'localhost' === wp_parse_url( $client_id, PHP_URL_HOST ) ) {
 			return;
 		}
 
@@ -38,10 +41,10 @@ class IndieAuth_Client_Discovery {
 
 	public function export() {
 		return array(
-			'manifest'    => $this->manifest,
 			'rels'        => $this->rels,
 			'mf2'         => $this->mf2,
 			'html'        => $this->html,
+			'json'        => $this->json,
 			'client_id'   => $this->client_id,
 			'client_name' => $this->client_name,
 			'client_icon' => $this->client_icon,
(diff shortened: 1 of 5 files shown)


dshanske:
@pfefferle Can you have a look?

janboddez:
Ha, missed this PR the other day. I'll see if I can try it tonight!

dshanske:
Feedback is appreciated

includes/class-indieauth-client-discovery.php
```diff
@@ -74,45 +75,63 @@ private function parse( $url ) {
...
-			if ( array_key_exists( 'logo', $this->mf2 ) ) {
-				if ( is_string( $this->mf2['logo'][0] ) ) {
-					$this->client_icon = $this->mf2['logo'][0];
-				} else {
-					$this->client_icon = $this->mf2['logo'][0]['value'];
+		} elseif ( 'text/html' === $content_type ) {
+			$content = wp_remote_retrieve_body( $response );
+
+			if ( class_exists( 'Masterminds\\HTML5' ) ) {
+				$domdocument = new \Masterminds\HTML5( array( 'disable_html_ns' => true ) );
```
artlung: Should this library be installed with composer and so be in the `composer.json` file?

includes/class-indieauth-client-discovery.php
```diff
@@ -74,45 +75,63 @@ private function parse( $url ) {
...
-
-		if ( class_exists( 'Masterminds\\HTML5' ) ) {
-			$domdocument = new \Masterminds\HTML5( array( 'disable_html_ns' => true ) );
-			$domdocument = $domdocument->loadHTML( $content );
-		} else {
-			$domdocument = new DOMDocument();
-			libxml_use_internal_errors( true );
-			if ( function_exists( 'mb_convert_encoding' ) ) {
-				$content = mb_convert_encoding( $content, 'HTML-ENTITIES', mb_detect_encoding( $content ) );
+		$content_type = wp_remote_retrieve_header( $response, 'content-type' );
```
artlung, line 83: `parse()` seems like it should be designated as `static` since that's how it is used. Also it seems as though it could get to the bottom and not `return` so I think it needs a return at the bottom. I also suggest a doc block describe the params and what it does if there's an error.

includes/class-indieauth-client-discovery.php
```diff
@@ -74,45 +75,63 @@ private function parse( $url ) {
...
+		if ( 'application/json' === $content_type ) {
+			$json = json_decode( wp_remote_retrieve_body( $response ), true );
```
artlung:
```suggestion
			/** @var $json array{
			 *  client_id: string,
			 *     client_name: string,
			 *     logo_uri: string
			 * }
			 */
			$json = json_decode( wp_remote_retrieve_body( $response ), true );
```

artlung: I like using comments to describe what we expect to be in json like this since it makes identifying and using them slightly easier.

includes/class-indieauth-client-discovery.php
```diff
@@ -5,6 +5,7 @@ class IndieAuth_Client_Discovery {
 	protected $manifest = array();
 	protected $html     = array();
 	protected $mf2      = array();
+	protected $json     = array();
```
artlung, line 7: You define this but then don't use it. Required?

includes/class-indieauth-client-discovery.php
```diff
@@ -74,45 +75,63 @@ private function parse( $url ) {
...
-			$domdocument = $domdocument->loadHTML( $content );
-		} else {
-			$domdocument = new DOMDocument();
-			libxml_use_internal_errors( true );
-			if ( function_exists( 'mb_convert_encoding' ) ) {
-				$content = mb_convert_encoding( $content, 'HTML-ENTITIES', mb_detect_encoding( $content ) );
+		$content_type = wp_remote_retrieve_header( $response, 'content-type' );
+		if ( 'application/json' === $content_type ) {
+			$json = json_decode( wp_remote_retrieve_body( $response ), true );
+			if ( ! is_array( $json ) && empty( $json ) ) {
```
janboddez: Guessing maybe this should be `|| empty( $json )` ? Like, you wanna return when it's either empty _or_ not an array (in other words, it has to be both an array and non-empty).

janboddez:
So, the PR works!

Also (I can't add a comment because it isn't new), this here:
```
private function get_html( $input ) {
	if ( ! $input ) {
		return;
	}
```

`$input` is always an instance of `DOMDocument` (or `Masterminds\\HTML5`), so you will never return here. Not really a problem, but it doesn't really add anything. I think.

Also, the issue I had earlier was caused because of:
```
$this->html['title'] = $xpath->query( '//title' )->item( 0 )->textContent;
```
If the response contains no `title` (rare, for HTML documents, but not impossible), you'd get at least a warning here. (Though somehow I was getting an infinite loop.) Better assign the outcome of `$xpath->query( '//title' )` to a variable and then loop over it (and `break` after the first iteration). Or run it through `if ( ! empty() )` or similar.

Not sure what happens if no meaningful client details are discovered? I liked Aaron's suggestion (I think) of just showing a URL ...?

includes/class-indieauth-client-discovery.php
dshanske: Removed it. It is an alternative to what we have, but it isn't necessary for what we're doing.

includes/class-indieauth-client-discovery.php
dshanske: Fixed that. It is meant to hold the json for returning.

PR #272, opened 2024-08-04
Check if download is valid before trying to sideload it

dshanske:
This adds a simple check to a sideload function that failed.


pfefferle:
I think we have to change the unittest framework first, because 6.6.1 does no longer support PHP < 7.2

dshanske:
I know, I figured that should be a separate pr

pfefferle (approved):
👍

Issue #265, opened 2024-05-18
Permission Error with Jetpack


ChaosExAnima:
It appears that there's a conflict with this plugin and the popular [Jetpack](https://github.com/automattic/jetpack) plugin. When this plugin is activated, attempting to access the Jetpack dashboard results in a "You are not allowed to access this page" error, even as an admin.

Steps to reproduce (I used [Studio](https://developer.wordpress.com/studio/) for a clean site):

1. Install IndieAuth and the Jetpack plugin.
2. Activate Jetpack.
3. Visit the Jetpack dashboard page, verifying no permission issues.
4. Activate the IndieAuth plugin.
5. Retry 3.

Expected: Jetpack dashboard is accessible.
Actual: Jetpack dashboard shows WordPress error "Sorry, you are not allowed to access this page".

I took a quick look at the code, but wasn't able to spot anything that was obviously the culprit. There's nothing in the PHP logs either.

dshanske:
@ChaosExAnima Will look into this. Don't use Jetpack myself, but assume it is our login code interfering with theirs. It shouldn't, but...

dshanske:
@akirk Worried this may affect the same code you've been submitting PRs for.

Kadams64:
I am seeing the same problem re: Jetpack and IndieAuth.  JetPack features fail with "You are not allowed to access this page" If IndieAuth is active
- Wordpress 6.6.1
- IndieAuth 4.5.0

I skimmed the [JetPack 'Known issues' page](https://jetpack.com/support/getting-started-with-jetpack/known-issues/) and saw this reference about [conflicts with plugins that change or restrict access to the WordPress REST API](https://jetpack.com/support/getting-started-with-jetpack/known-issues/#plugins-that-restrict-access-to-the-rest-api) that might (?) be relevant.

dshanske:
This doesn't do that...I am going to have to dig into what might be causing it. Unless Jetpack doesn't like sharing the connection...

dshanske:
@Kadams64 Do you want to try the update and see if you are satisfied?

Kadams64:
> @Kadams64 Do you want to try the update and see if you are satisfied?

It appears to solve the Jetpack conflict!  

I downloaded the current trunk and activated it, then tested each of the Jetpack left nav links: the permissions issue with "You are not allowed to access this page" no longer occurs and the expected Jetpack pages render.  
[...]

PR #273, opened 2024-08-12
Update class-indieauth-authorize.php

janboddez:
Even though the default value for `$error` _is_ `null`, we should probably not make assumptions about expected _return values_. By returning `$error` rather than a hardcoded `null`, we respect other plugins' possible changes.

Jetpack, for instance, calls this filter like so (i.e., with `$error` equal to `true`): https://github.com/Automattic/jetpack/blob/e29cee2d08516e93486e4dcd0b102244e7609ed4/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php#L190-L195. May be weird, but if we then return `null`, things break.


janboddez:
Oh, forgot to mention that this should fix #265.

janboddez:
> May be weird, but if we then return `null`, things break.

Turns out `true` is actually valid ... (https://developer.wordpress.org/reference/hooks/rest_authentication_errors/). Either way, probably safest to just allow other plugins to do their thing and just return whatever value we were passed (or an error, of course)!

pfefferle (approved):
👍

janboddez:
Made a few more changes.

Just one, actually, the rest was updating the PHPDoc.

Anyway, so, I was wondering. If the `$error` arg is in fact non-empty (because the filter was called, like Jetpack does, with `true` as its first argument), should we _always_ simply return it?

Maybe it's equally "safe" to still allow IndieAuth to return any error it encountered? Or rather, if this isn't an "IndieAuth request," no error would get set anyway (at least, that's what I'm assuming), and as a result the `$error` arg would be returned unchanged regardless. If that's the case, there's no need for the "empty" check. (I think?)

dshanske:
Looks good to me.

PR #280, opened 2024-12-30
Add Variable Analysis and Fix Results

dshanske:
This adds some additional tests and fixes some undeclared variables using the new testing.


dshanske:
@pfefferle Could you have a look at the amendments to the testing protocol here? I'd like to deploy this to other IndieWeb plugins.

Issue #286, opened 2025-02-01
Introspection could use SUB claim for Pure OAuth client Compatibility


carrvo:
For the [introspection endpoint](https://github.com/indieweb/wordpress-indieauth/blob/4a4aaf62221978c101bdd6900e9f1779ea97ea48/includes/class-indieauth-introspection-endpoint.php#L69), it only returns the `active` claim in the identity token but [the spec requires more](https://indieauth.spec.indieweb.org/#access-token-verification-response) (and so does the underlying OAuth spec).

More specifically, [mod_oauth2](https://github.com/OpenIDC/mod_oauth2) requires at least the `sub` claim. I believe this is so that it can set the user in Apache for other modules to process.

carrvo:
I am mistaken. It does return more claims in the identity token.
Additionally, the `sub` claim is neither required for IndieAuth nor OAuth.
However, the `sub` claim is required to be compatible with [mod_oauth2](https://github.com/OpenIDC/mod_oauth2) (a pure OAuth client for Apache).

I would be grateful if this were added; but can be satisfied with it closed without fix.

carrvo:
Evidence that required fields are implemented can be seen [in the test](https://github.com/indieweb/wordpress-indieauth/blob/7b163ab671d7e889e54532a39b7cd1d4f7396769/tests/test-introspection-endpoint.php#L100).

PR #288, opened 2025-02-14
SUB claim for Pure OAuth client Compatibility

carrvo:
the sub claim is required to be compatible with [mod_oauth2](https://github.com/OpenIDC/mod_oauth2) (a pure OAuth client for Apache).

This is the minimal implementation of https://github.com/indieweb/wordpress-indieauth/issues/286.

### DIFF
--- a/includes/class-token-user.php
+++ b/includes/class-token-user.php
@@ -190,6 +190,7 @@ class Token_User extends Token_Generic {
 		}
 
 		$value['user'] = $user_id;
+		$value['sub'] = $value['me'];
 		return $value;
 	}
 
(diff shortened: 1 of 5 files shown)


carrvo:
If it is desired to have authentication/authorization implemented (which the spec does require), then I would recommend doing so in another issue and pull request.

dshanske (changes requested):
Tests failed

carrvo:
My apologies, I am not particularly familiar with PHP unit testing yet...

pfefferle:
Added a safety check to only set the `sub` claim when the `me` property is present. This prevents potential PHP warnings if `me` is not set in the token data.

PR #291, opened 2025-10-22
Fix CSRF vulnerability in authorization endpoint

pfefferle:
Adds nonce verification to the authorization endpoint and includes nonce fields in authentication and authorization forms to enhance CSRF protection. Updates plugin version to 4.5.5 and documents the security fix in the changelog.


dshanske (approved):
Worked in testing

PR #295, opened 2025-10-23
Remove Ability to Avoid PKCE

dshanske:
PKCE has been a required security feature of IndieAuth for 5 years. Making its usage mandatory in this plugin for security reasons. No more fallback on less secure.

PKCE prevents CSRF and authorization code injection attacks.

Suggested alternative or supplement to #291 

It seems if PKCE solves this problem, then there is no need to use the WordPress nonce system. I suppose there is no issue in double checks, but...

### DIFF
--- a/includes/class-indieauth-authorization-endpoint.php
+++ b/includes/class-indieauth-authorization-endpoint.php
@@ -99,12 +99,12 @@ class IndieAuth_Authorization_Endpoint extends IndieAuth_Endpoint {
 						/* Code Challenge.
 						 */
 						'code_challenge'        => array(
-							'required' => true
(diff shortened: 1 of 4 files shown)


pfefferle (approved):
Can you add it to the changelog

lostfocus:
Hm. Any ideas how to log into https://github.com/aaronpk/Compass then? It doesn't have PKCE yet. (And with the current state of IndieAuth it seems to be one of the few tools/websites that I'm using IndieAuth on as well.)

dshanske:
@aaronpk Would you be willing to update Compass for PKCE?

PR #298, opened 2025-12-20
Add sub claim for OAuth client compatibility

### LINKED ISSUE #286: Introspection could use SUB claim for Pure OAuth client Compatibility

pfefferle:
## Summary

Adds the `sub` claim to token responses, setting it to the same value as `me`. This is required for compatibility with pure OAuth clients like mod_oauth2.

- Only sets `sub` when `me` property is present (safety check)
- Updates tests to include `me` in test tokens and expect `sub` in responses

Fixes #286

Based on #288 by @carrvo

### DIFF
--- a/includes/class-token-user.php
+++ b/includes/class-token-user.php
@@ -190,6 +190,9 @@ class Token_User extends Token_Generic {
 		}
 
 		$value['user'] = $user_id;
+		if ( isset( $value['me'] ) ) {
+			$value['sub'] = $value['me'];
+		}
 		return $value;
 	}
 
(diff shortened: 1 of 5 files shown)


pfefferle:
Sorry, I accidentally merged it! I reverted everything and re-added the PR.

dshanske:
The only reason I was reluctant about this is that it technically is not in the spec.

PR #314, opened 2026-07-06
Fix fatal error from undefined function __e()

### LINKED ISSUE #313: PHP Fatal error:  Uncaught Error: Call to undefined function __e() in indieauth-authenticate-form.php
I just had this error message in my logs after trying to log into a website with my domain. I was logged out of WordPress at the time.

I'll look into it myself, I won't be able to do that until Tuesday evening, though.

pfefferle:
## Summary

Fixes #313

`templates/indieauth-authenticate-form.php` called `__e()` on its error path — a function that does not exist in WordPress (typo of `esc_html_e()`, introduced in d04401a). Whenever the authorization form was rendered and `get_url_from_user()` returned no URL for the current user (e.g. the reporter's logged-out scenario), the page died with:

```
PHP Fatal error: Uncaught Error: Call to undefined function __e()
```

## Changes

- Replace `__e()` with `esc_html_e()` so the intended error message is actually displayed.
- Call `login_footer()` before `exit` so the login page HTML is properly closed, matching the other templates.

## Testing

- `php -l` and `phpcs` pass on the template.
- Verified no other `__e(` calls exist in the codebase.

### DIFF
--- a/templates/indieauth-authenticate-form.php
+++ b/templates/indieauth-authenticate-form.php
@@ -8,8 +8,8 @@ login_header(
 );
 $user_website = esc_url( get_url_from_user( $current_user->ID ) );
 if ( ! $user_website ) {
-	esc_html_e( 'The application cannot sign you in as WordPress cannot determine the current user', 'indieauth' );
-	login_footer();
+	\esc_html_e( 'The application cannot sign you in as WordPress cannot determine the current user', 'indieauth' );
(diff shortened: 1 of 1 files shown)


lostfocus:
That would do it 😄 I just opened my editor, `_e(` instead of `__e(` would have done the trick as well, I guess.

PR #326, opened 2026-08-06
harden pkce parameter validation

pfefferle:
Folding this into #322 instead, closing.


pfefferle:
Folding into #322, no reason to carry a separate branch for this.

